PR #29, opened 2019-10-02
Runtime scale implementation for Service Bus queues/topics

### LINKED ISSUE #30: Update Service Bus extension to support Scale Monitor Interface
Part of Azure/azure-functions-host#4821 

The service bus extension needs to implement IScaleMonitor in order to support the new runtime scale monitoring feature of premium functions.

gzuber:
Resolves #30

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusEntityPathHelper.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusEntityPathHelper.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    internal static class ServiceBusEntityPathHelper
+    {
+        public static EntityType ParseEntityType(string entityPath)
+        {
+            return entityPath.IndexOf("/Subscriptions/", StringComparison.OrdinalIgnoreCase) >= 0 ? EntityType.Topic : EntityType.Queue;
+        }
+
+        public static void ParseTopicAndSubscription(string entityPath, out string topic, out string subscription)
+        {
+            string[] arr = Regex.Split(entityPath, "/Subscriptions/", RegexOptions.IgnoreCase);
+
+            if (arr.Length < 2)
+            {
+                throw new InvalidOperationException($"{entityPath} is either formatted incorrectly, or is not a valid Service Bus subscription path");
+            }
+
+            topic = arr[0];
+            subscription = arr[1];
+        }
+    }
+}
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -2,19 +2,27 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Azure.ServiceBus;
 using Microsoft.Azure.ServiceBus.Core;
+using Microsoft.Azure.ServiceBus.Management;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
-    internal sealed class ServiceBusListener : IListener
+    internal sealed class ServiceBusListener : IListener, IScaleMonitorProvider
     {
         private readonly MessagingProvider _messagingProvider;
+        private readonly string _functionId;
+        private readonly EntityType _entityType;
         private readonly string _entityPath;
         private readonly bool _isSessionsEnabled;
         private readonly ServiceBusTriggerExecutor _triggerExecutor;
@@ -22,7 +30,9 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
         private readonly MessageProcessor _messageProcessor;
         private readonly ServiceBusAccount _serviceBusAccount;
         private readonly ServiceBusOptions _serviceBusOptions;
-        private MessageReceiver _receiver;
+        private readonly ILoggerFactory _loggerFactory;
+
+        private Lazy<MessageReceiver> _receiver;
         private ClientEntity _clientEntity;
         private bool _disposed;
         private bool _started;
@@ -30,18 +40,26 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
         private IMessageSession _messageSession;
         private SessionMessageProcessor _sessionMessageProcessor;
 
-        public ServiceBusListener(string entityPath, bool isSessionsEnabled, ServiceBusTriggerExecutor triggerExecutor, ServiceBusOptions config, ServiceBusAccount serviceBusAccount, MessagingProvider messagingProvider)
+        private Lazy<ServiceBusScaleMonitor> _scaleMonitor;
+
+        public ServiceBusListener(string functionId, EntityType entityType, string entityPath, bool isSessionsEnabled, ServiceBusTriggerExecutor triggerExecutor,
+            ServiceBusOptions config, ServiceBusAccount serviceBusAccount, MessagingProvider messagingProvider, ILoggerFactory loggerFactory)
         {
+            _functionId = functionId;
+            _entityType = entityType;
             _entityPath = entityPath;
             _isSessionsEnabled = isSessionsEnabled;
             _triggerExecutor = triggerExecutor;
             _cancellationTokenSource = new CancellationTokenSource();
             _messagingProvider = messagingProvider;
             _serviceBusAccount = serviceBusAccount;
+            _loggerFactory = loggerFactory;
+            _receiver = CreateMessageReceiver();
+            _scaleMonitor = new Lazy<ServiceBusScaleMonitor>(() => new ServiceBusScaleMonitor(_functionId, _entityType, _entityPath, _serviceBusAccount.ConnectionString, _receiver, _loggerFactory));
 
             if (_isSessionsEnabled)
             {
-                _sessionMessageProcessor = _messagingProvider.CreateSessionMessageProcessor(entityPath, _serviceBusAccount.ConnectionString);
+                _sessionMessageProcessor = _messagingProvider.CreateSessionMessageProcessor(_entityPath, _serviceBusAccount.ConnectionString);
             }
             else
             {
@@ -50,7 +68,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             _serviceBusOptions = config;
         }
 
-        internal MessageReceiver Receiver => _receiver;
+        internal MessageReceiver Receiver => _receiver.Value;
 
         internal IMessageSession MessageSession => _messageSession;
 
@@ -79,8 +97,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             }
             else
             {
-                _receiver = _messagingProvider.CreateMessageReceiver(_entityPath, _serviceBusAccount.ConnectionString);
-                _receiver.RegisterMessageHandler(ProcessMessageAsync, _serviceBusOptions.MessageHandlerOptions);
+                Receiver.RegisterMessageHandler(ProcessMessageAsync, _serviceBusOptions.MessageHandlerOptions);
             }
             _started = true;
 
@@ -100,10 +117,10 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             // ProcessMessageAsync invocations to cancel
             _cancellationTokenSource.Cancel();
 
-            if (_receiver != null)
+            if (_receiver != null && _receiver.IsValueCreated)
             {
-                await _receiver.CloseAsync();
-                _receiver = null;
+                await Receiver.CloseAsync();
+                _receiver = CreateMessageReceiver();
             }
             if (_clientEntity != null)
             {
@@ -130,9 +147,9 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                 // For now, rely on finalization to clean up _cancellationTokenSource's wait handle (if allocated).
                 _cancellationTokenSource.Cancel();
 
-                if (_receiver != null)
+                if (_receiver != null && _receiver.IsValueCreated)
                 {
-                    _receiver.CloseAsync().Wait();
+                    Receiver.CloseAsync().Wait();
                     _receiver = null;
                 }
 
@@ -146,6 +163,10 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             }
         }
 
+        private Lazy<MessageReceiver> CreateMessageReceiver()
+        {
+            return new Lazy<MessageReceiver>(() => _messagingProvider.CreateMessageReceiver(_entityPath, _serviceBusAccount.ConnectionString));
+        }
 
         internal async Task ProcessMessageAsync(Message message, CancellationToken cancellationToken)
         {
@@ -176,6 +197,11 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             {
                 throw new ObjectDisposedException(null);
             }
+        }
+
+        public IScaleMonitor GetMonitor()
+        {
+            return _scaleMonitor.Value;
         }
     }
 }
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
@@ -1,0 +1,314 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.Azure.ServiceBus;
+using Microsoft.Azure.ServiceBus.Core;
+using Microsoft.Azure.ServiceBus.Management;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    internal class ServiceBusScaleMonitor : IScaleMonitor<ServiceBusTriggerMetrics>
+    {
+        private const string DeadLetterQueuePath = @"/$DeadLetterQueue";
+
+        private readonly string _functionId;
+        private readonly EntityType _entityType;
+        private readonly string _entityPath;
+        private readonly string _connectionString;
+        private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
+        private readonly bool _isListeningOnDeadLetterQueue;
+        private readonly Lazy<MessageReceiver> _receiver;
+        private readonly Lazy<ManagementClient> _managementClient;
+        private readonly ILogger<ServiceBusScaleMonitor> _logger;
+
+        private DateTime _nextWarningTime;
+
+        public ServiceBusScaleMonitor(string functionId, EntityType entityType, string entityPath, string connectionString, Lazy<MessageReceiver> receiver, ILoggerFactory loggerFactory)
+        {
+            _functionId = functionId;
+            _entityType = entityType;
+            _entityPath = entityPath;
+            _connectionString = connectionString;
+            _scaleMonitorDescriptor = new ScaleMonitorDescriptor($"{_functionId}-ServiceBusTrigger-{_entityPath}".ToLower());
+            _isListeningOnDeadLetterQueue = entityPath.EndsWith(DeadLetterQueuePath);
+            _receiver = receiver;
+            _managementClient = new Lazy<ManagementClient>(() => new ManagementClient(_connectionString));
+            _logger = loggerFactory.CreateLogger<ServiceBusScaleMonitor>();
+            _nextWarningTime = DateTime.UtcNow;
+        }
+
+        public ScaleMonitorDescriptor Descriptor
+        {
+            get
+            {
+                return _scaleMonitorDescriptor;
+            }
+        }
+
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
+        {
+            return await GetMetricsAsync();
+        }
+
+        public async Task<ServiceBusTriggerMetrics> GetMetricsAsync()
+        {
+            Message message = null;
+            string entityName = _entityType == EntityType.Queue ? "queue" : "topic";
+
+            try
+            {
+                // Peek the first message in the queue without removing it from the queue
+                message = await _receiver.Value.PeekAsync();
+
+                if (_entityType == EntityType.Queue)
+                {
+                    return await GetQueueMetricsAsync(message);
+                }
+                else
+                {
+                    return await GetTopicMetricsAsync(message);
+                }
+            }
+            catch (MessagingEntityNotFoundException)
+            {
+                _logger.LogWarning($"ServiceBus {entityName} '{_entityPath}' was not found.");
+            }
+            catch (UnauthorizedException) // When manage claim is not used on Service Bus connection string
+            {
+                if (TimeToLogWarning())
+                {
+                    _logger.LogWarning($"Connection string does not have Manage claim for {entityName} '{_entityPath}'. Failed to get {entityName} description to " +
+                        $"derive {entityName} length metrics. Falling back to using first message enqueued time.");
+                }
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning($"Error querying for Service Bus {entityName} scale status: {e.Message}");
+            }
+
+            // Path for connection strings with no manage claim
+            return CreateTriggerMetrics(message, 0, 0, 0, _isListeningOnDeadLetterQueue);
+        }
+
+        private async Task<ServiceBusTriggerMetrics> GetQueueMetricsAsync(Message message)
+        {
+            QueueRuntimeInfo queueRuntimeInfo;
+            QueueDescription queueDescription;
+            long activeMessageCount = 0, deadLetterCount = 0;
+            int partitionCount = 0;
+
+            queueRuntimeInfo = await _managementClient.Value.GetQueueRuntimeInfoAsync(_entityPath);
+            activeMessageCount = queueRuntimeInfo.MessageCountDetails.ActiveMessageCount;
+            deadLetterCount = queueRuntimeInfo.MessageCountDetails.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
+            queueDescription = await _managementClient.Value.GetQueueAsync(_entityPath);
+            partitionCount = queueDescription.EnablePartitioning ? 16 : 0;
+
+            return CreateTriggerMetrics(message, activeMessageCount, deadLetterCount, partitionCount, _isListeningOnDeadLetterQueue);
+        }
+
+        private async Task<ServiceBusTriggerMetrics> GetTopicMetricsAsync(Message message)
+        {
+            TopicDescription topicDescription;
+            SubscriptionRuntimeInfo subscriptionRuntimeInfo;
+            string topicPath, subscriptionPath;
+            long activeMessageCount = 0, deadLetterCount = 0;
+            int partitionCount = 0;
+
+            ServiceBusEntityPathHelper.ParseTopicAndSubscription(_entityPath, out topicPath, out subscriptionPath);
+
+            subscriptionRuntimeInfo = await _managementClient.Value.GetSubscriptionRuntimeInfoAsync(topicPath, subscriptionPath);
+            activeMessageCount = subscriptionRuntimeInfo.MessageCountDetails.ActiveMessageCount;
+            deadLetterCount = subscriptionRuntimeInfo.MessageCountDetails.DeadLetterMessageCount;
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
+            topicDescription = await _managementClient.Value.GetTopicAsync(topicPath);
+            partitionCount = topicDescription.EnablePartitioning ? 16 : 0;
+
+            return CreateTriggerMetrics(message, activeMessageCount, deadLetterCount, partitionCount, _isListeningOnDeadLetterQueue);
+        }
+
+        internal static ServiceBusTriggerMetrics CreateTriggerMetrics(Message message, long activeMessageCount, long deadLetterCount, int partitionCount, bool isListeningOnDeadLetterQueue)
+        {
+            long totalNewMessageCount = 0;
+            TimeSpan queueTime = TimeSpan.Zero;
+
+            if (message != null)
+            {
+                queueTime = DateTime.UtcNow.Subtract(message.SystemProperties.EnqueuedTimeUtc);
+                totalNewMessageCount = 1; // There's at least one if message != null. Default for connection string with no manage claim
+            }
+
+            if ((!isListeningOnDeadLetterQueue && activeMessageCount > 0) || (isListeningOnDeadLetterQueue && deadLetterCount > 0))
+            {
+                totalNewMessageCount = isListeningOnDeadLetterQueue ? deadLetterCount : activeMessageCount;
+            }
+
+            return new ServiceBusTriggerMetrics
+            {
+                MessageCount = totalNewMessageCount,
+                PartitionCount = partitionCount,
+                QueueTime = queueTime
+            };
+        }
+
+        ScaleStatus IScaleMonitor.GetScaleStatus(ScaleStatusContext context)
+        {
+            return GetScaleStatusCore(context.WorkerCount, context.Metrics?.Cast<ServiceBusTriggerMetrics>().ToArray());
+        }
+
+        public ScaleStatus GetScaleStatus(ScaleStatusContext<ServiceBusTriggerMetrics> context)
+        {
+            return GetScaleStatusCore(context.WorkerCount, context.Metrics?.ToArray());
+        }
+
+        private ScaleStatus GetScaleStatusCore(int workerCount, ServiceBusTriggerMetrics[] metrics)
+        {
+            ScaleStatus status = new ScaleStatus
+            {
+                Vote = ScaleVote.None
+            };
+
+            const int NumberOfSamplesToConsider = 5;
+
+            // Unable to determine the correct vote with no metrics.
+            if (metrics == null || metrics.Length == 0)
+            {
+                return status;
+            }
+
+            // We shouldn't assign more workers than there are partitions
+            // This check is first, because it is independent of load or number of samples.
+            int partitionCount = metrics.Last().PartitionCount;
+            if (partitionCount > 0 && partitionCount < workerCount)
+            {
+                status.Vote = ScaleVote.ScaleIn;
+                _logger.LogInformation($"WorkerCount ({workerCount}) > PartitionCount ({partitionCount}).");
+                _logger.LogInformation($"Number of instances ({workerCount}) is too high relative to number " +
+                                       $"of partitions for Service Bus entity ({_entityPath}, {partitionCount}).");
+                return status;
+            }
+
+            // At least 5 samples are required to make a scale decision for the rest of the checks.
+            if (metrics.Length < NumberOfSamplesToConsider)
+            {
+                return status;
+            }
+
+            // Maintain a minimum ratio of 1 worker per 1,000 messages.
+            long latestMessageCount = metrics.Last().MessageCount;
+            if (latestMessageCount > workerCount * 1000)
+            {
+                status.Vote = ScaleVote.ScaleOut;
+                _logger.LogInformation($"MessageCount ({latestMessageCount}) > WorkerCount ({workerCount}) * 1,000.");
+                _logger.LogInformation($"Message count for Service Bus Entity ({_entityPath}, {latestMessageCount}) " +
+                                       $"is too high relative to the number of instances ({workerCount}).");
+                return status;
+            }
+
+            // Check to see if the queue/topic has been empty for a while. Only if all metrics samples are empty do we scale down.
+            bool isIdle = metrics.All(m => m.MessageCount == 0);
+            if (isIdle)
+            {
+                status.Vote = ScaleVote.ScaleIn;
+                _logger.LogInformation($"'{_entityPath}' is idle.");
+                return status;
+            }
+
+            // Samples are in chronological order. Check for a continuous increase in message count.
+            // If detected, this results in an automatic scale out for the site container.
+            if (metrics[0].MessageCount > 0)
+            {
+                bool messageCountIncreasing =
+                IsTrueForLastN(
+                    metrics,
+                    NumberOfSamplesToConsider,
+                    (prev, next) => prev.MessageCount < next.MessageCount) && metrics[0].MessageCount > 0;
+                if (messageCountIncreasing)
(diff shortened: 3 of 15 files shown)


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
                 throw new ObjectDisposedException(null);
             }
         }
+
+        private void InitializeMetricsClients()
```
gzuber: Using this as a way to avoid spending the resources creating these clients if the listener isn't going to be using runtime scale.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+                }
+
+                isUnauthorized = true;
+            }
+            catch (Exception e)
+            {
+                _logger.LogWarning($"Error querying for Service Bus queue scale status: {e.Message}");
+            }
+
+            return CompileTriggerMetrics(message, activeMessageCount, deadLetterCount, partitionCount, isUnauthorized, _isListeningOnDeadLetterQueue);
```
gzuber: CompileTriggerMetrics is being used partially to avoid code duplication between queues and topics and partially to allow for testing of that logic.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            DateTime currentTime = DateTime.UtcNow;
+            bool timeToLog = currentTime >= _nextWarningTime;
+            if (timeToLog)
+            {
+                _nextWarningTime = currentTime.AddHours(1);
+            }
+            return timeToLog;
+        }
+
+        private void ParseTopicAndSubscription(string entityPath, out string topic, out string subscription)
```
gzuber: This is the thing I like the least about this implementation. I have to have topic and subscription separated to get the subscription runtime. I wasn't able to find a way to do this in the SDK (only ways to combine them). I did find [this](https://github.com/Azure/azure-service-bus-dotnet/issues/67) github thread that makes it seem like doing it this way would work, but it would break if it ever changed. The only time we have this info separated and available is in `ServiceBusTriggerAttributeBindingProvider` but I thought it would be weird to keep passing that down through the tree, especially since the repo is trying to make these classes generic to work for topics or queues

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -2,42 +2,65 @@
...
 {
-    internal sealed class ServiceBusListener : IListener
+    internal sealed class ServiceBusListener : IListener, IScaleMonitor<ServiceBusTriggerMetrics>
     {
-        private readonly MessagingProvider _messagingProvider;
+        private const string DeadLetterQueuePath = @"/$DeadLetterQueue";
+
+        private readonly MessagingProvider _messagingProvider;
+        private readonly string _functionId;
+        private readonly EntityType _entityType;
```
mathewc: @alrod had just went through and removed this EntityType I believe

gzuber: Yeah so I noticed the listener had recently been changed to be generic, but the issue is the GetMetricsAsync code changes based on whether it's a queue or a topic so the listener needs to know what EntityType it is. If there's a better way to do it I'm all ears. I thought of potentially treating everything as a queue and switching to topic if it fails, or looking at the entityPath for "/subscription/", but this seemed cleaner to me at the time

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -2,42 +2,65 @@
...
         private readonly ServiceBusAccount _serviceBusAccount;
-        private readonly ServiceBusOptions _serviceBusOptions;
-        private MessageReceiver _receiver;
-        private ClientEntity _clientEntity;
+        private readonly ServiceBusOptions _serviceBusOptions;
+        private readonly ScaleMonitorDescriptor _scaleMonitorDescriptor;
+        private readonly ILogger<ServiceBusListener> _logger;
+
+        private MessageReceiver _receiver;
+        private MessageReceiver _metricsReceiver;
```
mathewc: Do we need a separate receiver for metrics? If we change the existing Receiver property to do lazy create, we could reuse it for both purposes (both metrics and main listener)?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: How about just using properties with lazy init?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            }
+            else
+            {
+                return await GetTopicMetricsAsync();
+            }
+        }
+
+        public async Task<ServiceBusTriggerMetrics> GetMetricsAsync()
+        {
+            if (_entityType == EntityType.Queue)
```
mathewc: Let's not duplicate this code

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            else
+            {
+                return await GetTopicMetricsAsync();
+            }
+        }
+
+        public async Task<ServiceBusTriggerMetrics> GetQueueMetricsAsync()
+        {
+            Message message = null;
+            QueueDescription queue;
```
mathewc: nit: queueDescription

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            QueueRuntimeInfo queueRuntime;
```
mathewc: nit: queueRuntimeInfo

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: nit: maybe name CreateTriggerMetrics?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            int partitionCount = 0;
+            TimeSpan eventQueueTime = TimeSpan.Zero;
+            bool isUnauthorized = false;
+
+            try
+            {
+                InitializeMetricsClients();
+
+                // Peek the first message in the queue without removing it from the queue
+                message = await _metricsReceiver.PeekAsync();
```
alrod: Did you test it with sessions enabled entity? My understanding MessageReceiver does not work with sessions.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            if (!isUnauthorized && ((!isListeningOnDeadLetterQueue && activeMessageCount > 0) || (isListeningOnDeadLetterQueue && deadLetterCount > 0)))
+            {
+                totalNewEventCount = isListeningOnDeadLetterQueue ? deadLetterCount : activeMessageCount;
+            }
+
+            return new ServiceBusTriggerMetrics
+            {
+                Timestamp = DateTime.UtcNow,
+                TotalNewEventCount = totalNewEventCount,
+                PartitionCount = partitionCount,
```
alrod: Why do we need `int` if it's always `16` or `0`? Maybe bool property?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: I am ok to bring back EntityType but parsing entityPath looks also ok to me.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
                 throw new ObjectDisposedException(null);
             }
         }
+
+        private void InitializeMetricsClients()
+        {
+            if (_metricsReceiver == null)
+            {
+                _metricsReceiver = _messagingProvider.CreateMessageReceiver(_entityPath, _serviceBusAccount.ConnectionString);
```
alrod: `_metricsReceiver` will be equal `_receiver` (if sessions is not enabled). Why do we need both?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            }
+
+            if (_managementClient == null)
+            {
+                _managementClient = new ManagementClient(_serviceBusAccount.ConnectionString);
+            }
+        }
+
+        async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()
```
alrod: `async Task<ScaleMetrics> IScaleMonitor.GetMetricsAsync()` and public `async Task<ServiceBusTriggerMetrics> GetMetricsAsync()` have the same body. Can we create a private method:
`return _entityType == EntityType.Queue ? await GetQueueMetricsAsync() : await GetTopicMetricsAsync();`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: Yeah, you're right. The listener was generalized such that it no longer had queue/topic specific operations in it, hidden behind MessagingProvider.CreateClientEntity for the most part. And CreateClientEntity has similar parsing code in it already

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+
+                // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+                // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
+                queue = await _managementClient.GetQueueAsync(_entityPath);
+                if (queue.EnablePartitioning)
+                {
+                    partitionCount = 16;
+                }
+            }
+            catch (MessagingEntityNotFoundException)
```
alrod: Exceptions handling is the same for `GetQueueMetricsAsync` and `GetTopicMetricsAsync`. Can we avoid copy/paste somehow?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            {
+                return await GetQueueMetricsAsync();
+            }
+            else
+            {
+                return await GetTopicMetricsAsync();
+            }
+        }
+
+        public async Task<ServiceBusTriggerMetrics> GetQueueMetricsAsync()
```
alrod: Why `GetQueueMetricsAsync` and `GetTopicMetricsAsync` are public?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            if (anchor < 0)
+            {
+                throw new InvalidOperationException($"{entityPath} is either formatted incorectly, or is not a valid Service Bus subscription path");
+            }
+
+            topic = entityPath.Substring(0, anchor);
+            subscription = entityPath.Substring(anchor + divider.Length, entityPath.Length - (anchor + divider.Length));
+        }
+
+        private static bool IsTrueForLastN(IList<ServiceBusTriggerMetrics> samples, int count, Func<ServiceBusTriggerMetrics, ServiceBusTriggerMetrics, bool> predicate)
```
mathewc: I realize the various implementations are repeating this helper :) I though about adding it to the SDK as a helper, but didn't want to expand the surface area unecessarily. We might add helpers like this in the future if they are common.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+            return new ServiceBusTriggerMetrics
+            {
+                Timestamp = DateTime.UtcNow,
+                TotalNewEventCount = totalNewEventCount,
+                PartitionCount = partitionCount,
+                EventQueueTime = eventQueueTime
+            };
+        }
+
+        ScaleStatus IScaleMonitor.GetScaleStatus(ScaleStatusContext context)
```
alrod: `private` ?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -2,42 +2,65 @@
...
+using Microsoft.Azure.ServiceBus.Management;
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using Microsoft.Azure.WebJobs.Host.Scale;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
-    internal sealed class ServiceBusListener : IListener
+    internal sealed class ServiceBusListener : IListener, IScaleMonitor<ServiceBusTriggerMetrics>
```
mathewc: Since this ServiceBus implementation is more complicated than some of the others, one option to keep the listener code simple would be to move all the monitor stuff into its own ServiceBusScaleMonitor class, and have this listener implement IScaleMonitorProvider and return an instance (as BlobListener does [here](https://github.com/Azure/azure-webjobs-sdk/blob/2c7b2f30b4d197d091f5aa5543ddf65d6a34e689/src/Microsoft.Azure.WebJobs.Extensions.Storage/Blobs/Listeners/BlobListener.cs#L91)).

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+        {
+            ScaleStatus status = new ScaleStatus
+            {
+                Vote = ScaleVote.None
+            };
+
+            const int NumberOfSamplesToConsider = 5;
+
+            // Unable to determine the correct vote with no metrics.
+            if (metrics == null)
```
alrod: Is possible to get empty array here? It will broke the code below if yes.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: Can we add some unit tests for `GetScaleStatus` methods?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+                throw new InvalidOperationException($"{entityPath} is either formatted incorectly, or is not a valid Service Bus subscription path");
+            }
+
+            topic = entityPath.Substring(0, anchor);
+            subscription = entityPath.Substring(anchor + divider.Length, entityPath.Length - (anchor + divider.Length));
+        }
+
+        private static bool IsTrueForLastN(IList<ServiceBusTriggerMetrics> samples, int count, Func<ServiceBusTriggerMetrics, ServiceBusTriggerMetrics, bool> predicate)
+        {
+            Debug.Assert(count > 1, "count must be greater than 1.");
```
alrod: Remove `Debug.Assert`s ?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +209,338 @@ private void ThrowIfDisposed()
...
+        {
+            return GetScaleStatusCore(context.WorkerCount, context.Metrics?.Cast<ServiceBusTriggerMetrics>().ToArray());
+        }
+
+        public ScaleStatus GetScaleStatus(ScaleStatusContext<ServiceBusTriggerMetrics> context)
+        {
+            return GetScaleStatusCore(context.WorkerCount, context.Metrics?.ToArray());
+        }
+
+        private ScaleStatus GetScaleStatusCore(int workerCount, ServiceBusTriggerMetrics[] metrics)
```
alrod: Is this code copy/paste from `ScaleController`?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
gzuber: So there is a section of the scaling logic (GetScaleStatusCore) that needs to use the actual number to make a decision. I can change the scaling logic to make an assumption about the number of partitions if that's better - I guess I just thought including the number in the metrics was cleaner and matches the other triggers

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
gzuber: I have about 9 right now in ServiceBusListenerTests with the prefix GetScaleStatus, am I missing a test case?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
gzuber: Yeah pretty much. I had to pick and choose which sections of the scale controller logic applied to service bus, but it all comes from ScaleController.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
gzuber: Ok I think I modified it but let me know if you had something else in mind

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -79,8 +94,7 @@ public Task StartAsync(CancellationToken cancellationToken)
             }
             else
             {
-                _receiver = _messagingProvider.CreateMessageReceiver(_entityPath, _serviceBusAccount.ConnectionString);
-                _receiver.RegisterMessageHandler(ProcessMessageAsync, _serviceBusOptions.MessageHandlerOptions);
+                _receiver.Value.RegisterMessageHandler(ProcessMessageAsync, _serviceBusOptions.MessageHandlerOptions);
```
mathewc: Can you change these call sites to use the Receiver property? Looks like it was done inconsistently before.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -177,5 +191,10 @@ private void ThrowIfDisposed()
                 throw new ObjectDisposedException(null);
             }
         }
+
+        public IScaleMonitor GetMonitor()
+        {
+            return new ServiceBusScaleMonitor(_functionId, _entityType, _entityPath, _serviceBusAccount.ConnectionString, _receiver, _loggerFactory);
```
mathewc: Let's not new one of these up each time we're called. Instead, can you either create it in the ctor, or define a field for it and create once/cache in this method?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Listeners/ServiceBusListenerTests.cs
```diff
@@ -33,20 +48,26 @@ public ServiceBusListenerTests()
...
+            _mockServiceBusAccount.Setup(a => a.ConnectionString).Returns(_testConnection);
+
+            _loggerFactory = new LoggerFactory();
+            _loggerProvider = new TestLoggerProvider();
+            _loggerFactory.AddProvider(_loggerProvider);
 
-            _listener = new ServiceBusListener(_entityPath, false, triggerExecutor, config, mockServiceBusAccount.Object, _mockMessagingProvider.Object);
+            _listener = new ServiceBusListener(_functionId, EntityType.Queue, _entityPath, false, _triggerExecutor, _serviceBusOptions, _mockServiceBusAccount.Object,
+                                _mockMessagingProvider.Object, _loggerFactory);
+            _scaleMonitor = (ServiceBusScaleMonitor)_listener.GetMonitor();
```
mathewc: Let's move all these new tests into a new ServiceBusScaleMonitorTests class, now that you've factored the logic cleanly out of the listener. In this file, you could just have one new test for the new GetMonitor method.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: Looks like in passing EntityType around you've made breaking changes to public interfaces. We need to not do that unfortunately

gzuber: Ok I reverted the interface changes, but changed it so both this and the ScaleMonitor are using the same logic to parse topic/subscription.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -100,9 +117,9 @@ public async Task StopAsync(CancellationToken cancellationToken)
             // ProcessMessageAsync invocations to cancel
             _cancellationTokenSource.Cancel();
 
-            if (_receiver != null)
+            if (Receiver != null)
```
mathewc: What will this do if the listener was never started and lazy receiver never evaluated? Will it end up creating it only to Close it?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.ServiceBus;
```
mathewc: nit: sort/order these usings. Perhaps your VS settings for this aren't configured correctly (system namespaces first)?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    class ServiceBusScaleMonitor : IScaleMonitor
```
mathewc: nit: internal class (explicit access modifier)

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
...
+    {
+        public ServiceBusMetricsData(Message message, long activeMessageCount, long deadLetterCount, int partitionCount)
+        {
+            Message = message;
+            ActiveMessageCount = activeMessageCount;
+            DeadLetterCount = deadLetterCount;
+            PartitionCount = partitionCount;
+        }
+
+        public Message Message { get; set; }
```
mathewc: This isn't part of the serializable metrics contract right? I.e. shouldn't be serialized to metrics table storage? If not, please remove, and instead pass the Message into your create metrics helper directly.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// <summary>
+        /// The number of partitions.
+        /// </summary>
+        public int PartitionCount { get; set; }
+
+        /// <summary>
+        /// The length of time the next event has been
+        /// sitting there.
+        /// </summary>
+        public TimeSpan EventQueueTime { get; set; }
```
mathewc: Rename to just QueueTime, since these are messages, not Events

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        public long TotalNewEventCount { get; set; }
+
+        /// <summary>
+        /// The number of partitions.
+        /// </summary>
+        public int PartitionCount { get; set; }
+
+        /// <summary>
+        /// The length of time the next event has been
```
mathewc: nit: "next message"

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host.Scale;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    class ServiceBusTriggerMetrics : ScaleMetrics
```
mathewc: internal class. Use explicit access modifiers for all classes

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// The number of new monitored events.
+        /// </summary>
+        public long TotalNewEventCount { get; set; }
```
mathewc: rename to MessageCount

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusTriggerMetrics.cs
```diff
@@ -0,0 +1,29 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    class ServiceBusTriggerMetrics : ScaleMetrics
+    {
+        /// <summary>
+        /// The number of new monitored events.
```
mathewc: nit: The number of messages currently in the queue/topic

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
...
+
+            const int NumberOfSamplesToConsider = 5;
+
+            // Unable to determine the correct vote with no metrics.
+            if (metrics == null)
+            {
+                return status;
+            }
+
+            // We shouldn't assign more workers than there are partitions (Cosmos DB, Event Hub, Service Bus Queue/Topic)
```
mathewc: This code is specific to ServiceBus so remove the generalized comments for Cosmos/EventHub

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
...
+            if (latestQueueLength > workerCount * 1000)
+            {
+                status.Vote = ScaleVote.ScaleOut;
+                _logger.LogInformation($"QueueLength ({latestQueueLength}) > WorkerCount ({workerCount}) * 1,000.");
+                _logger.LogInformation($"Length of queue for Service Bus Entity ({_entityPath}, {latestQueueLength}) " +
+                                       $"is too high relative to the number of instances ({workerCount}).");
+                return status;
+            }
+
+            // Check to see if the trigger source has been empty for a while. Only if all trigger sources are empty do we scale down.
```
mathewc: nit: rather than the general "trigger source", say "queue/topic"

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
...
+
+            // If partitioning is turned on, then Service Bus automatically partitions queues into 16 partitions
+            // See more information here: https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning#standard
+            topicDescription = await _managementClient.Value.GetTopicAsync(topicPath);
+            metrics.PartitionCount = topicDescription.EnablePartitioning ? 16 : 0;
+
+            return metrics;
+        }
+
+        public static ServiceBusTriggerMetrics CreateTriggerMetrics(ServiceBusMetricsData metrics, bool isUnauthorized, bool isListeningOnDeadLetterQueue)
```
mathewc: I'd just get rid of this ServiceBusMetricsData class, and instead pass the bits of info needed into this helper. Then GetQueueMetrics/GetTopicMetrics just call this helper and return a ScaleMetrics instance directly, w/o the need for an intermediate type.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
```diff
@@ -0,0 +1,337 @@
...
+            }
+
+            if (!isUnauthorized && ((!isListeningOnDeadLetterQueue && metrics.ActiveMessageCount > 0) || (isListeningOnDeadLetterQueue && metrics.DeadLetterCount > 0)))
+            {
+                totalNewEventCount = isListeningOnDeadLetterQueue ? metrics.DeadLetterCount : metrics.ActiveMessageCount;
+            }
+
+            return new ServiceBusTriggerMetrics
+            {
+                Timestamp = DateTime.UtcNow,
```
mathewc: nit: the default ctor will timestamp this, so you can remove this

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusEntityPathHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
+{
+    internal static class ServiceBusEntityPathHelper
+    {
+        public static EntityType ParseEntityType(string entityPath)
+        {
+            return entityPath.Contains("/Subscriptions/") ? EntityType.Topic : EntityType.Queue;
```
mathewc: We need these string operations to be case insensitive, here and other places. These aren't are they?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Bindings/ServiceBusTriggerAttributeBindingProviderTests.cs
```diff
@@ -4,11 +4,13 @@
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Castle.Core.Logging;
```
mathewc: I don't think this is needed?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Listeners/ServiceBusListenerTests.cs
```diff
@@ -89,10 +102,23 @@ public async Task ProcessMessageAsync_BeginProcessingReturnsFalse_MessageNotProc
...
+        public void GetMonitor_ReturnsExpectedValue()
+        {
+            IScaleMonitor scaleMonitor = _listener.GetMonitor();
+
+            Assert.Equal(scaleMonitor.GetType(), typeof(ServiceBusScaleMonitor));
+            Assert.Equal(scaleMonitor.Descriptor.Id, $"{_functionId}-ServiceBusTrigger-{_entityPath}".ToLower());
+
+            var scaleMonitor2 = _listener.GetMonitor();
+
+            Assert.Equal(scaleMonitor, scaleMonitor2);
```
mathewc: Assert.Same (instance equality, not value equality)

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Listeners/ServiceBusScaleMonitorTests.cs
```diff
@@ -0,0 +1,459 @@
...
+using Microsoft.Extensions.Options;
+using Moq;
+using Xunit;
+using static Microsoft.Azure.ServiceBus.Message;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus.UnitTests.Listeners
+{
+    public class ServiceBusScaleMonitorTests
+    {
+        //private readonly MessagingFactory _messagingFactory;
```
mathewc: remove

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Listeners/ServiceBusScaleMonitorTests.cs
```diff
@@ -0,0 +1,459 @@
...
+
+            _triggerExecutor = new ServiceBusTriggerExecutor(_mockExecutor.Object);
+            _mockServiceBusAccount = new Mock<ServiceBusAccount>(MockBehavior.Strict);
+            _mockServiceBusAccount.Setup(a => a.ConnectionString).Returns(_testConnection);
+
+            _loggerFactory = new LoggerFactory();
+            _loggerProvider = new TestLoggerProvider();
+            _loggerFactory.AddProvider(_loggerProvider);
+
+            _listener = new ServiceBusListener(_functionId, EntityType.Queue, _entityPath, false, _triggerExecutor, _serviceBusOptions, _mockServiceBusAccount.Object,
```
mathewc: these tests shouldn't need to create a listener just to create the monitor under test. Just create the monitor, and pass the mocks into it

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
gzuber: Yeah, that would be better. The only problem is that `CreateTriggerMetrics` needs the `isUnauthorized` flag. But we only get that flag in a catch block in `GetMetricsAsync`. I could move just the `isUnauthorized` catch to `GetQueueMetrics`/`GetTopicMetrics`, so we'd have the original try/catch split across those three functions, would that be cleaner than the intermediate type?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Listeners/ServiceBusScaleMonitorTests.cs
gzuber: So the issue here is that since the `MessageReceiver` is Lazy, by default it only calls `CreateMessageReceiver` once and caches the exception. So in this test it'll throw a `MessagingEntityNotFound` exception, cache it, and continue to throw that exception for the rest of the test. This is the workaround I came up with. The alternative is I could make it a `Lazy<MessageReceiver>` with `LazyThreadSafetyMode.PublicationOnly` so it doesn't cache exceptions. It doesn't look like this would cause any issues based on [these docs](https://docs.microsoft.com/en-us/dotnet/api/system.threading.lazythreadsafetymode?view=netframework-4.8) (assuming MessageReceiver is thread safe). Do you have any experience with this/an opinion?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
gzuber: I believe this should fix it, but would appreciate a second look

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -130,10 +147,10 @@ public void Dispose()
                 // For now, rely on finalization to clean up _cancellationTokenSource's wait handle (if allocated).
                 _cancellationTokenSource.Cancel();
 
-                if (_receiver != null)
+                if (_receiver.IsValueCreated)
                 {
-                    _receiver.CloseAsync().Wait();
-                    _receiver = null;
+                    Receiver.CloseAsync().Wait();
+                    _receiver = new Lazy<MessageReceiver>(() => _messagingProvider.CreateMessageReceiver(_entityPath, _serviceBusAccount.ConnectionString));
```
mathewc: On the Disposal path, just set this to null, and add a check for null above in both places "if (_receiver != null && _receiver.IsValueCreated)". Once disposed, the ThrowIfDisposed check will prevent start from being called again. I agree on the stop path with your reinitializing the Lazy receiver. Recommend moving the Lazy creation into a private helper CreateMessageReceiver so we're not duplicating the code

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: One minore comment on that, but otherwise OK

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
mathewc: I'm trying to understand why you even need the isUnauthorized bool. You're only using it to set the totalNewEventCount to zero, right? But if an exception is thrown trying to get the count, it'll be zero anyways since it was never set (exception from management client call)?

gzuber: So I re-examined the scale controller code and you're right, the isUnauthorized bool isn't necessary with the way we have it set up. I also caught a few bugs in my implementation testing the no manage claim scenario. Without a manage claim, we can peek a message but can't get info about the queue/runtime (active, dead, and partition counts). So we can peek the message in GetMetricsAsync and provide an alternate path to CreateTriggerMetrics when an Unauthroized exception is thrown in GetQueueMetrics/GetTopicMetrics. No managed claim should have MessageCount = 1 and QueueTime = some TimeSpan, which this should accomplish

PR #31, opened 2019-10-03
Apply  MaxConcurrentCalls depdnding on number of processors. Fixes #27

### LINKED ISSUE #27: Apply defaults based on environment capabilities and SKU
Currently, default configuration values are applied without taking the hosting environment into consideration. In order to correctly apply default configuration settings that take advantage of hosting capabilities (additional cores, memory, etc.), we need to take those factors into consideration.

This should apply to all SKUs, except consumption, where we need to ensure we maintain the same behavior we have today.

alrod:


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Config/ServiceBusOptions.cs
```diff
@@ -29,6 +29,13 @@ public ServiceBusOptions()
             };
 
             SessionHandlerOptions = new SessionHandlerOptions(ExceptionReceivedHandler);
+
+            // Set configuration for non-dynamic SKUs
+            string skuValue = Environment.GetEnvironmentVariable(Constants.AzureWebsiteSku);
+            if (!string.Equals(skuValue, Constants.DynamicSku, StringComparison.OrdinalIgnoreCase))
+            {
+                MessageHandlerOptions.MaxConcurrentCalls *= Environment.ProcessorCount;
+            }
```
brettsam: Are we doing this in response to an issue? Is this a recommendation from SB?

mathewc: Don't we have a tracking item for this that you can link to this bug? I think the work item across all triggers is to use more intelligent defaults across all our triggers, to improve performance.

What about SessionHandlerOptions? What is the default for PrefetchCount - should we also default that?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Config/ServiceBusOptionsTests.cs
```diff
@@ -28,7 +28,7 @@ public ServiceBusOptionsTests()
         public void Constructor_SetsExpectedDefaults()
         {
             ServiceBusOptions config = new ServiceBusOptions();
-            Assert.Equal(16, config.MessageHandlerOptions.MaxConcurrentCalls);
+            Assert.True(config.MessageHandlerOptions.MaxConcurrentCalls >= 16);
```
mathewc: Recommend making this check stronger and compute the expected value in the test based on processor count. For the E2E test as well.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Config/ServiceBusOptions.cs
alrod: tracking item is #27. `SessionHandlerOptions` doesn't have `PrefetchCount` but it have `MaxConcurrentSessions`. We don't want to change default `MaxConcurrentSessions` depending on core count.

mathewc: Ok

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Config/ServiceBusOptionsTests.cs
mathewc: Just fix these two test issues, and the code looks good to me

Issue #26, opened 2019-09-21
Processing stucks after getting SessionLockLostException


alrod:
Following code process a message only once using Microsoft.Azure.ServiceBus 3.4.0. (current package we are using in the extension) and then stcuks. The same code works fine using Microsoft.Azure.ServiceBus 4.0.0. Need to update to Microsoft.Azure.ServiceBus 4.0.0.

        static void Main(string[] args)
        {
            string cs = "";
            QueueClient client = new QueueClient(cs, "");

            SessionHandlerOptions options = new SessionHandlerOptions(ExceptionReceivedHandler)
            {
                MaxConcurrentSessions = 1,
                MaxAutoRenewDuration = TimeSpan.FromSeconds(30)
            };

            client.RegisterSessionHandler(ProcessSessionMessageAsync, options);
            Console.ReadLine();
        }

        static async Task ProcessSessionMessageAsync(IMessageSession session, Message message, CancellationToken cancellationToken)
        {
            Console.WriteLine($"Message delivary count: {message.SystemProperties.DeliveryCount}");
            await Task.Delay(50000);
        }

        static Task ExceptionReceivedHandler(ExceptionReceivedEventArgs args)
        {
            Console.WriteLine(args.Exception);
            return Task.CompletedTask;
        }

alrod:
Blocked on Azure/azure-sdk-for-net#6410

mpnow:
Hi @alrod, can you provide an update on what's happening with this bug? It's biting me on occasion in a production solution. Thanks!

alrod:
@mpnow, once ServiceBus releases the fix for Azure/azure-sdk-for-net#6410 I will merge PR to update ServiceBus SDK. Probably ServiceBus will release new ServiceBus SDK next week.

mpnow:
Hi @alrod, it looks like ServiceBus SDK 4.1.1 is out now - is that the release you were waiting for?

mpnow:
Hi @alrod, any news on getting this fix released? It's continuing to bite me periodically in a production app. Thanks!

mpnow:
Hi @alrod, wondering if a fix for this is out yet or, if not, when it might be?

alrod:
the fix is out, please use:
https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.ServiceBus/4.1.0

PR #32, opened 2019-10-04
Updating to Microsoft.Azure.ServiceBus 4.1.1. (#26)

### LINKED ISSUE #25: service bus binding does not accept EntityPath as connect string
_From @globysdz on May 26, 2017 18:48_

{
  "disabled": false,
  "bindings": [
    {
      "name" : "myFuncname",
      "topicName" : "request",
      "subscriptionName" : "processor",
      "connection" : "IncomingSubscription"
      "accessRights" : "listen",
      "type" : "serviceBusTrigger",
      "direction" : "in"
    }
  ]
}

the "incomingSubscription" was set via ARM template: 
    {
              "name": "IncomingSubscription",
              "value": "[listkeys(resourceId('Microsoft.ServiceBus/namespaces/topics/authorizationRules', parameters('bus_name'), 'request', 'perm-r'), variables('sbVersion')).primaryConnectionString]" 
     }

the result "inccomingSubscription" string contains ";EntityPath=xxxx" at the end of the string which AzureFunction throws exception at startup time:

Error indexing method 'Functions.globyscadProcessor' ---> System.ArgumentException : ConnectionString used to create MessagingFactory shouldn't include EntityPath as MessagingFactory is at Namespace level.

I am using the same connection string for my webapi application, which works fine. but for function it doesn't work. and there is no good way to automatically set the connection string variable (unless there is a way to obtain the connection string from ARM without EntityPath=xxx part).

_Copied from original issue: Azure/Azure-Functions#315_

alrod:
Merge this after the fix is released:
https://github.com/Azure/azure-sdk-for-net/issues/6410

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
@@ -6,7 +6,7 @@
     <RootNamespace>Microsoft.Azure.WebJobs.ServiceBus</RootNamespace>
     <PackageId>Microsoft.Azure.WebJobs.Extensions.ServiceBus</PackageId>
     <Description>Microsoft Azure WebJobs SDK ServiceBus Extension</Description>
-    <Version>3.2.0</Version>
+    <Version>4.0.0</Version>
     <CommitHash Condition="$(CommitHash) == ''">N/A</CommitHash>
     <InformationalVersion>$(Version) Commit hash: $(CommitHash)</InformationalVersion>
     <Authors>Microsoft</Authors>
@@ -37,7 +37,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="3.4.0" />
+    <PackageReference Include="Microsoft.Azure.ServiceBus" Version="4.1.1" />
     <PackageReference Include="Microsoft.Azure.WebJobs" Version="3.0.14" />
     <PackageReference Include="Microsoft.Azure.WebJobs.Sources" Version="3.0.10" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta004">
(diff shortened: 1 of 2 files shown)


Message.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="us-ascii"?>
+<Message xmlns="http://schemas.microsoft.com/servicebusexplorer">
+  <Date>Friday, October 4, 2019 10:14:45 PM</Date>
+  <Content><![CDATA[<message>Hi mate, how are you?</message>]]></Content>
+</Message>
```
brettsam: What are these Message.xml and Properties.xml files?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
```diff
@@ -235,6 +235,8 @@ private async Task ServiceBusEndToEndInternal<T>(IHost host = null)
                 string[] consoleOutputLines = consoleOutput
                     .Where(p => p.FormattedMessage != null)
                     .SelectMany(p => p.FormattedMessage.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries))
+                    // There is a known issue in SB SDK 4.0.0 in MessageReceivePump.MessagePumpTaskAsync https://github.com/Azure/azure-sdk-for-net/issues/6410
+                    .Where(p => !p.StartsWith("Message processing error"))
```
brettsam: How prevalent is this error? Will this cause a lot of errors logged for customers (and thus CRIs)? Looks like they've fixed it -- should we wait for their next release to move forward?

Message.xml
alrod: We do not need them, add to PR by mistake. I will remove them.

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
alrod: Make swnse. I will delay the release.

Issue #15, opened 2017-02-15
Service Bus Batch Trigger


BrianVallelunga:
I'd like to start a conversation around supporting batches of messages from an Azure Service Bus. A similar request was made for Storage Queues in https://github.com/Azure/azure-webjobs-sdk/issues/625

Ideally, I'd be able to specify something like:

    public void Handle([ServiceBusBatchTrigger("myqueue", 25)] BrokeredMessage[] messages)
    {
      // Process the batch of messages  
    }

There are a lot of scenarios where batching message processing is far more efficient than working with an individual message. In my specific case, I'm doing database inserts and those could be done much more efficiently in a batch.

I'm going to look into a similar project that supports Storage Queues: https://github.com/ealsur/WebJobs.Extensions.GroupQueueTrigger

Is this something others would be interested in?

SeanFeldman:
@alrod could you please share a link to the documentation (or PR) for this new addition? Thank you.

alrod:
There is not documentation in the moment.
Using:
```
        [FunctionName("Function1")]
        public static void Run([ServiceBusTrigger("myqueue", Connection = "AzureWebJobsServiceBus")]string[] mySbMsg, ILogger log)
        {..}
```
Also you can specify `Message[], byte[][], PocoClass[]` as the input trigger param.

Note `SingletonOptions` defined in `host.json` don't apply to batch processing

PR:
https://github.com/Azure/azure-functions-servicebus-extension/pull/11

nzthiago:
@alrod I *think* when you use ISessionClient (which inherits and expands on IMessageReceiver) it will still be able to ReceiveAsync with a batch for a specific SessionId (gotta call AcceptMessageSessionAsync first to get an IMessageSession). 

I created a simple sample netcoreapp2.2 here: https://github.com/nzthiago/ServiceBusBatchReceiveWithSessions

As you can see it sends 10 messages to 2 sessions ("1", and "2"). Then it tries to read from only session "1" in batches of 5 and it's successful: 
![image](https://user-images.githubusercontent.com/1563288/63902326-ed5fef00-c9bd-11e9-9877-fdbd10236c11.png)

rasavant-ms:
Would be nice to have the metadata properties available for the earlier versions supported as valid parameter Types in the new package version - Microsoft.Azure.WebJobs.Extensions.ServiceBus 3.1.0-beta4-batch

[...]

PR #39, opened 2019-11-20
Batch support

alrod:
Addresses https://github.com/Azure/azure-functions-servicebus-extension/issues/15


jeffhollan:
Thanks @alrod - excited to see this feature getting shipped.  I can’t speak much to the PR itself but mostly just stating what I’d expect for this type of “batch” or “single”approach that we have for triggers like Event Hubs.  From skimming the code I think this covers it but a few items you could likely confirm:

- In C# if I set the type as `string`, `Message`, `byte[]` I would only get a single execution
- In C# if I set the type as `string[]`, `Message[]`, or `byte[][]`(? I don’t know what we expect for a collection of byte arrays in Event Hubs actually) it would automatically switch to batch mode.  The batch would be ordered for sessions.
- In non-.NET if I configure the function.json with a property like `cardinality` = `many` then it will pass in the batch the the out-of-proc worker.  My trigger on that type would be something like `string[]` as well (whatever we have in the Event Hub templates by default should be cardinality many).  Same question though on how we do a `byte[][]`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+            if (_receiver == null)
+            {
+                _receiver = CreateMessageReceiver();
+            }
+
+            if (_isSessionsEnabled)
+            {
+                _clientEntity = _messagingProvider.CreateClientEntity(_entityPath, _serviceBusAccount.ConnectionString);
+
+                if (_singleDispatch)
```
mathewc: Looks like when singleDispatch is false you do the same thing here, regardless of whether sessions is enabled. So it'd be good to hoist this out to it's own, e.g.

```csharp
if (_singleDispatch)
{
    if (_sessionsEnabled)
    {
    }
    else
    {
    }
}
else
{
}
```

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+            }
+
+            // cancel our token source to signal any in progress
+            // ProcessMessageAsync invocations to cancel
+            _cancellationTokenSource.Cancel();
+
+            if (_receiver != null && _receiver.IsValueCreated)
+            {
+                await Receiver.CloseAsync();
+                _receiver = null;
```
mathewc: @gzuber made the code this way in his [recent commit](https://github.com/Azure/azure-functions-servicebus-extension/commit/abc92e91bdc02b36a2e6ceae8f0b03ee65e747dd), I think to ensure the Lazy was initialized?

gzuber: Oh it looks like @alrod added the `CreateMessageReceiver` call to `StartAsync` so that should work (making the assumption that nothing can be called between a Stop and subsequent Start)

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+                receiver = Receiver;
+            }
+
+            Task.Run(async () =>
+            {
+                while (true)
+                {
+                    try
+                    {
+                        if (_receiver == null || cancellationToken.IsCancellationRequested)
```
mathewc: Related to the Dispose comment above. Previously _recevier wouldn't be null anymore - it was a Lazy. So _receiver.IsValueCreated was used. If we're reverting all that, we need to make sure we're doing it consistently everywhere

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+                                receiver = await sessionClient.AcceptMessageSessionAsync();
+                            }
+                            catch (ServiceBusTimeoutException)
+                            {
+                                // it's expected if the entity is empty, try next time
+                                continue;
+                            }
+                        }
+
+                        IList<Message> messages = messages = await receiver.ReceiveAsync(_serviceBusOptions.BatchOptions.MaxMessageCount, _serviceBusOptions.BatchOptions.OperationTimeout);
```
mathewc: What happens if the operation timeout expires? Is an exception thrown, or are we just waiting until either MaxMessages are available, otherwise we return 0 messages?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+
+                        if (messages != null)
+                        {
+                            Message[] messagesArray = messages.ToArray();
+                            ServiceBusTriggerInput input = new ServiceBusTriggerInput()
```
mathewc: For singleton, you've used the ServiceBusTriggerInput.New pattern, but here you're using a ctor. Why don't you just go all in on the create pattern, and create to factory methods: CreateSingle and CreateBatch, and make the ctor private?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Azure.ServiceBus.Core;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    // The core object we get when an ServiceBus is triggered. 
+    // This gets converted to the user type (Message, string, poco, etc) 
+    internal sealed class ServiceBusTriggerInput
+    {
+        // If != -1, then only process a single message in this batch. 
+        private int _selector = -1;
```
mathewc: This whole int selector thing is weird and we should remove it. I realize it's a copied pattern from EventHub. It's only used in the GetSingleMessage method, and that method should go away (see comments there).

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs
```diff
@@ -0,0 +1,58 @@
...
+            return new ServiceBusTriggerInput
+            {
+                Messages = this.Messages,
+                MessageReceiver = this.MessageReceiver,
+                MessageSession = this.MessageSession,
+                _selector = idx
+            };
+        }
+
+        public Message GetSingleMessage()
```
mathewc: We should just remove this method, and make the Messages array property public. Then it's up to the consumer to use messages as needed based on singleton/batch mode.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        // If != -1, then only process a single message in this batch. 
+        private int _selector = -1;
+
+        public MessageReceiver MessageReceiver { get; set; }
+        public IMessageSession MessageSession { get; set; }
+
+        internal Message[] Messages { get; set; }
+
+        public bool IsSingleDispatch
```
mathewc: Based on which factory method is called (CreateSingle/CreateBatch) you'd just set this appropriately. No need for a selector.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+                            {
+                                Messages = messagesArray
+                            };
+                            if (_isSessionsEnabled)
+                            {
+                                input.MessageSession = receiver as IMessageSession;
```
mathewc: Rather than these two MessageSession/MessageReceiver properties, I think you should just have a single IMessageReceiver property. Consumer can then check its type as needed and use appropriately?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+        {
+            if (!await _messageProcessor.BeginProcessingMessageAsync(message, cancellationToken))
+            {
+                return;
+            }
+
+            ServiceBusTriggerInput input = ServiceBusTriggerInput.New(message);
+            input.MessageReceiver = Receiver;
+
+            FunctionResult result = await _triggerExecutor.TryExecuteAsync(GetTriggerFunctionData(message, input), cancellationToken);
```
mathewc: nit: put the triggered function data on its own line, assigned to its own var (for this and other occurrences below).

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+                            }
+
+                            FunctionResult result = await _triggerExecutor.TryExecuteAsync(GetTriggerFunctionData(messagesArray, input), cancellationToken);
+
+                            if (cancellationToken.IsCancellationRequested)
+                            {
+                                return;
+                            }
+
+                            // Complete batch of messages only if the execution was succesfull
```
mathewc: nit: spelling

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+        {
+            SessionClient sessionClient = null;
+            IMessageReceiver receiver = null;
+            if (_isSessionsEnabled)
+            {
+                sessionClient = _messagingProvider.CreateSessionClient(_entityPath, _serviceBusAccount.ConnectionString);
+            }
+            else
+            {
+                receiver = Receiver;
```
mathewc: For the receiver created here on either path, are all the AutoComplete/AutoRenew/etc. options in effect? How do those options affect batch processing?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+
+                            if (cancellationToken.IsCancellationRequested)
+                            {
+                                return;
+                            }
+
+                            // Complete batch of messages only if the execution was succesfull
+                            if (result.Succeeded && Receiver != null)
+                            {
+                                await receiver.CompleteAsync(messagesArray.Select(x => x.SystemProperties.LockToken));
```
mathewc: So the contract with the function is that unless it throws an exception, we'll assume all messages were handled successfully and we'll complete them all, otherwise we don't complete any, and they pop back into the queue after their locks expire? How is $dead-letter handled in this case - will the messages be dead lettered after N attempts as normal?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+                            }
+                        }
+                    }
+                    catch (ObjectDisposedException)
+                    {
+                        // Ignore as we are stopping the host
+                    }
+                    catch (NullReferenceException)
+                    {
+                        // Ignore as we are stopping the host
```
mathewc: nit: if you get a nullref you don't know it's because the host is stopping. I don't think we should be swallowing nullrefs

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+                    }
+                    catch (Exception ex)
+                    {
+                        // Log another exception
+                        _logger.LogError($"Exception has occurred during batch listening:: {ex.ToString()}");
```
mathewc: nit: "An unhandled exception occurred in the message batch receive loop: {ex.ToString()}"

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+            }
+
+            ServiceBusTriggerInput input = ServiceBusTriggerInput.New(message);
+            input.MessageSession = session;
+
+            FunctionResult result = await _triggerExecutor.TryExecuteAsync(GetTriggerFunctionData(message, input), cancellationToken);
+            await _sessionMessageProcessor.CompleteProcessingMessageAsync(session, message, result, cancellationToken);
+        }
+
+        internal void StartMultipleMessageListening(CancellationToken cancellationToken)
```
mathewc: suggest rename to StartMessageBatchReceiver

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: What if an exception occurs that is fatal, meaning listening is no longer happening? Can that happen? Wondering if this listener can get in to a zombie failed state.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+
         public IScaleMonitor GetMonitor()
         {
             return _scaleMonitor.Value;
-        }
-    }
-}
+        }
+
+        private TriggeredFunctionData GetTriggerFunctionData(Message message, ServiceBusTriggerInput input)
```
mathewc: Recommend just defining a ToTriggerFunctionData method on ServiceBusTriggerInput, to keep all these details encapsulated. Internally it'll know the dispatch mode and can do the right thing.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+            DateTime[] enqueuedTimes = new DateTime[length];
+            DateTime[] lockedUntils = new DateTime[length];
+            string[] sessionIds = new string[length];
+            for (int i = 0; i < messages.Length; i++)
+            {
+                messageIds[i] = messages[i].MessageId;
+                deliveryCounts[i] = messages[i].SystemProperties.DeliveryCount;
+                enqueuedTimes[i] = messages[i].SystemProperties.EnqueuedTimeUtc;
+                lockedUntils[i] = messages[i].SystemProperties.LockedUntilUtc;
+                sessionIds[i] = messages[i].SessionId;
```
mathewc: Won't there still only be 1 session? Can a batch include messages from different sessions?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,375 @@
...
+            return new TriggeredFunctionData()
+            {
+                ParentId = parentId,
+                TriggerValue = input,
+                TriggerDetails = new Dictionary<string, string>()
+                {
+                    { "MessageIds", string.Join(",", messageIds)},
+                    { "DeliveryCounts", string.Join(",", deliveryCounts) },
+                    { "EnqueuedTimes", string.Join(",", enqueuedTimes) },
+                    { "LockedUntils", string.Join(",", lockedUntils) },
```
mathewc: to avoid the issue with strange names, recommend you just append "Array" to the end of each, as we did in EventGrid binding data for batches [here](https://github.com/Azure/azure-functions-eventhubs-extension/blob/7ccd930e6d2fdde64ded0ec7f540b08627739619/src/Microsoft.Azure.WebJobs.Extensions.EventHubs/Triggers/EventHubTriggerInputBindingStrategy.cs#L102)

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/MessageToStringConverter.cs
```diff
@@ -1,72 +0,0 @@
...
-                try
-                {
-                    return await reader.ReadToEndAsync();
-                }
-                catch (DecoderFallbackException)
-                {
-                    // we'll try again below
-                }
-
-                // We may get here if the message is a string yet was DataContract-serialized when created. We'll
```
mathewc: The new Message->String converter you added is just a one liner. How does it differ from what we were doing before - are you sure there are no breaking changes here?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerBindingStrategy.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.Azure.ServiceBus;
+using Microsoft.Azure.ServiceBus.Core;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using Microsoft.Azure.WebJobs.Host.Triggers;
+
+namespace Microsoft.Azure.WebJobs.ServiceBus
+{
+    // Binding strategy for an service bus triggers.
```
mathewc: nit: fix wording

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: 1. It returns `null` messages if there are no new messages during "timeout"
2. If there are new message it returns immediately with `messages.Count <= MaxMessages`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: `AutoComplete/AutoRenew` are affected only single processing as the options can be passed to  `RegisterMessageHandler/RegisterSessionHandler`. `ReceiveAsync` does not use the options.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: I moved back `_receiver = CreateMessageReceiver();` in `StopAsync`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: we can check `_started` here instead of `_receiver.IsValueCreated`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: Without sessions failed messages automatically go to dead letter queue after N attempts. In sessions case I added the code to abandon messages:
  
Service Bus SDK (`IMessageReceiver.ReceiveAsync`) does nor support `MessageHandlerOptions.AutoComplete` so we do not provide ability to users to complete messages themselves.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: The listener should not go to "zombie  failed  state". Listening is happen on each loop step. I will monitor unhandled exceptions after the feature release.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: You are right. There is only one session.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/MessageToStringConverter.cs
alrod: It shouldn't break, I added tests for different conversions.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,325 @@
...
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            ThrowIfDisposed();
+
+            if (_started)
+            {
+                throw new InvalidOperationException("The listener has already been started.");
+            }
+
+
```
mathewc: fix formatting

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -1,207 +1,325 @@
...
+                    }
+                }
+                else
+                {
+                    Receiver.RegisterMessageHandler(ProcessMessageAsync, _serviceBusOptions.MessageHandlerOptions);
+                }
+            }
+            else
+            {
+                StartMessageBatchReceiver(_cancellationTokenSource.Token);
```
mathewc: It looks like in the Sessions support PR from earlier this year, you removed the code we had that flowed tokens from this source down into ProcessMessage (v2 code [here](https://github.com/Azure/azure-webjobs-sdk/blob/a31126d8d2bff90e30c972b718a1b7ea42fefe81/src/Microsoft.Azure.WebJobs.ServiceBus/Listeners/ServiceBusListener.cs#L111)). You need to add that back, so cancellation works as expected.

ericsampson:
Just wanted to say that I’m looking forward to this, thanks @alrod and @mathewc 👍🏼

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: Got it. So how will customers ensure their message locks are retained if they aren't being renewed? If they're in the middle of a long running function processing a large batch of messages, and the locks time out, another host may start processing these messages again concurrently.

Similarly for AutoComplete - what if the user wants/needs to manage complete/abandon/deadletter operations themselves for the messages.

Both of the above issues have come up a ton over the years in this extension.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
mathewc: Right - MessageHandlerOptions.AutoComplete is only used when you're registering a message handler with the receiver. It's assumed that if you're using the receive APIs yourself as you're doing here, that you handle the completions yourself. So we will need to allow users to do that. Why can't we just honor whatever value they've configured for MessageHandlerOptions.AutoComplete and only do the complete/abandon if set to true?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: `RegisterMessageHandler` requires the function:
`Func<Message, CancellationToken, Task>` handler

so I created linked `CancellationToken` to listen both tokens

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs
```diff
@@ -0,0 +1,115 @@
...
+                Message message = Messages[0];
+                if (IsSingleDispatch)
+                {
+                    Guid? parentId = ServiceBusCausalityHelper.GetOwner(message);
+                    return new TriggeredFunctionData()
+                    {
+                        ParentId = parentId,
+                        TriggerValue = this,
+                        TriggerDetails = new Dictionary<string, string>()
+                {
```
mathewc: nit: fix formatting

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs
```diff
@@ -0,0 +1,115 @@
...
+                        enqueuedTimes[i] = Messages[i].SystemProperties.EnqueuedTimeUtc;
+                        lockedUntils[i] = Messages[i].SystemProperties.LockedUntilUtc;
+                    }
+
+                    return new TriggeredFunctionData()
+                    {
+                        ParentId = parentId,
+                        TriggerValue = this,
+                        TriggerDetails = new Dictionary<string, string>()
+                {
```
mathewc: nit: fix formatting

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusTriggerStrategyTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            Assert.Equal(messages.Length, ((string[])bindingData["ContentTypeArray"]).Length);
+            Assert.Equal(messages.Length, ((string[])bindingData["ReplyToArray"]).Length);
+            Assert.Equal(messages.Length, ((long[])bindingData["SequenceNumberArray"]).Length);
+            Assert.Equal(messages.Length, ((string[])bindingData["ToArray"]).Length);
+            Assert.Equal(messages.Length, ((string[])bindingData["LabelArray"]).Length);
+            Assert.Equal(messages.Length, ((string[])bindingData["CorrelationIdArray"]).Length);
+            Assert.Equal(messages.Length, ((IDictionary<string, object>[])bindingData["UserPropertiesArray"]).Length);
+        }
+
+        private static void CheckBindingContract(Dictionary<string, Type> bindingDataContract)
```
mathewc: nit: move all these private helpers after all the tests

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusTriggerStrategyTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            TestHelpers.SetField(sysProps, "lockedUntilUtc", DateTime.MinValue);
+            TestHelpers.SetField(sysProps, "sequenceNumber", 1);
+            TestHelpers.SetField(sysProps, "enqueuedTimeUtc", DateTime.MinValue);
+            TestHelpers.SetField(sysProps, "lockTokenGuid", Guid.NewGuid());
+            TestHelpers.SetField(sysProps, "deadLetterSource", "test");
+            return sysProps;
+        }
+
+        [Fact]
+        public void TriggerStrategy()
```
mathewc: Give this test a better name

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Triggers/ServiceBusTriggerBindingIntegrationTests.cs
```diff
@@ -1,90 +0,0 @@
...
-            _queueBinding = new ServiceBusTriggerBinding("parameterName", typeof(UserDataType), argumentBinding, null, options, messagingProvider, EntityType.Queue, "queueName", false, NullLoggerFactory.Instance);
-            _topicBinding = new ServiceBusTriggerBinding("parameterName", typeof(UserDataType), argumentBinding, null, options, messagingProvider, EntityType.Topic, "subscriptionName/Subsciptions/topicName", false, NullLoggerFactory.Instance);
-        }
-
-        [Theory]
-        [InlineData("RequestId", "4b957741-c22e-471d-9f0f-e1e8534b9cb6")]
-        [InlineData("RequestReceivedTime", "8/16/2014 12:09:36 AM")]
-        [InlineData("DeliveryCount", "8")]
-        [InlineData("IsSuccess", "False")]
-        public void BindAsync_IfUserDataType_ReturnsValidBindingData(string userPropertyName, string userPropertyValue)
```
mathewc: Do we have tests that verify poco binding e2e?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusSessionsEndToEndTests.cs
```diff
@@ -376,6 +435,79 @@ public class ServiceBusSessionsTestJobs2
...
+                }
+
+                if (firstReceived && secondReceived)
+                {
+                    _waitHandle1.Set();
+                }
+            }
+        }
+
+        public class ServiceBusMultipleMessagesTestJobs1
```
mathewc: nit: Give all these test classes better names, e.g. ServiceBusMultipleMessagesTestJob_BindToStringArray, ServiceBusMultipleMessagesTestJob_BindToMessageArray, ServiceBusMultipleMessagesTestJob_BindToPocoArray.

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
```diff
@@ -505,4 +685,10 @@ public void Dispose()
             Cleanup().GetAwaiter().GetResult();
         }
     }
+
+    public class DummyClass
```
mathewc: nit: TestPoco :)

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
```diff
@@ -248,16 +309,16 @@ private async Task ServiceBusEndToEndInternal<T>(IHost host = null)
                     $"{jobContainerType.FullName}.SBTopicListener2",
                     $"{jobContainerType.FullName}.ServiceBusBinderTest",
                     "Job host started",
-                    $"Executing '{jobContainerType.Name}.SBQueue2SBQueue' (Reason='New ServiceBus message detected on '{FirstQueueName}'.', Id=",
```
mathewc: This is an unfortunate loss. EventHub doesn't have a Reason either? Assume this is because the "strategy" pattern doesn't allow you to specify your own ParameterDescriptor, which means you can't implement GetTriggerReason

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/MessageToStringConverterTests.cs
```diff
@@ -1,92 +0,0 @@
...
-
-namespace Microsoft.Azure.WebJobs.ServiceBus.UnitTests
-{
-    public class MessageToStringConverterTests
-    {
-        private const string TestString = "This is a test!";
-        private const string TestJson = "{ value: 'This is a test!' }";
-
-        [Theory]
-        [InlineData(ContentTypes.TextPlain, TestString)]
```
mathewc: You should retain these tests, running the data through your new converter logic. That way we're sure we don't have any regressions. These tests were added in the past to guard against regressions.

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Listeners/ServiceBusListenerTests.cs
```diff
@@ -1,107 +1,110 @@
...
+            typeof(Message.SystemPropertiesCollection).GetProperty("EnqueuedTimeUtc").SetValue(systemProperties, DateTime.Now);
+            typeof(Message.SystemPropertiesCollection).GetProperty("LockedUntilUtc").SetValue(systemProperties, DateTime.Now);
+            typeof(Message).GetProperty("SystemProperties").SetValue(message, systemProperties);
+
+            message.MessageId = Guid.NewGuid().ToString();
+            CancellationToken cancellationToken = new CancellationToken();
+            _mockMessageProcessor.Setup(p => p.BeginProcessingMessageAsync(message, cancellationToken)).ReturnsAsync(true);
+
+            FunctionResult result = new FunctionResult(true);
+            _mockExecutor.Setup(p => p.TryExecuteAsync(It.Is<TriggeredFunctionData>(q => (q.TriggerValue as ServiceBusTriggerInput).Messages[0] == message), cancellationToken)).ReturnsAsync(result);
```
mathewc: nit: use direct cast rather than "as" because you expect a particular type

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Bindings/ServiceBusTriggerBindingTests.cs
```diff
@@ -1,101 +0,0 @@
...
-using Moq;
-
-namespace Microsoft.Azure.WebJobs.ServiceBus.UnitTests.Bindings
-{
-    public class ServiceBusTriggerBindingTests
-    {
-        [Fact]
-        public void CreateBindingDataContract_ReturnsExpectedValue()
-        {
-            IReadOnlyDictionary<string, Type> argumentContract = null;
```
mathewc: I don't see where you retained these tests?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
alrod: Yes, EventHub does not have a Reason. Cosmos trigger creates `CosmosDBTriggerBinding`, ServiceBus/EventHub create `ITriggerBinding` using `BindingFactory.GetTriggerBinding(new ServiceBusTriggerBindingStrategy(), parameter, _converterManager, createListener)`.

we can add passing descriptor to `BindingFactory.GetTriggerBinding`. But it will require SDK a change,

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/Bindings/ServiceBusTriggerBindingTests.cs
alrod: there are no such thing as thing `ServiceBusTriggerBinding`, Now it's `ServiceBusTriggerBindingStrategy`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Config/BatchOptions.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        /// <summary>
+        /// The time span the client waits for receiving a message before it times out.
+        /// </summary>
+        public TimeSpan OperationTimeout { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value that indicates whether the messages should be completed after successful processing. 
+        /// </summary>
+        public bool AutoComplete { get; set; }
```
nzthiago: Does this mean that AutoComplete will be able to be set to False by devs creating Azure Functions with the service bus trigger? If so it'd be great to include tests with AutoComplete = false in this PR. Wondering how would a dev complete the messages inside a function, especially for non .NET languages that don't bind to Message...

Issue #54, opened 2020-01-23
Uncaught MessageLockLostException when a single message is dead lettered from a message batch


elksson:
Steps to reproduce:

`        [FunctionName("Reproduction")]
        public async Task Reproduction(
    [ServiceBusTrigger("topic", "subscription", Connection = "ServiceBusConnection")] long[] servicetypes,
    Int32[] deliveryCountArray,
    string[] lockTokenArray,
    MessageReceiver messageReceiver,
    IBinder binder,
    ILogger log)
        {
            await messageReceiver.DeadLetterAsync(lockTokenArray[0]);
        }`

alrod:
@elksson, did you set AutoCompete to false?
https://github.com/Azure/azure-functions-servicebus-extension/blob/dev/test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs#L356

This will skip completing in the code

elksson:
@alrod I did not set autocomplete to false cause I did want the auto complete to happen if no messages needed to be dead lettered. That would be a workable solution however 99% of the time I am expecting that the function will run without any errors so I won't need to dead letter any messages and would like the trigger to handle completing them.

elksson:
@alrod when autocomplete is enabled for single mode instead of batch does the exception get thrown if the lock is lost before autocomplete? or does the run loop swallow this exception?

alrod:
> @alrod I did not set autocomplete to false cause I did want the auto complete to happen if no messages needed to be dead lettered. That would be a workable solution however 99% of the time I am expecting that the function will run without any errors so I won't need to dead letter any messages and would like the trigger to handle completing them.

If you want to use `DeadLetterAsync` in your code you need to set AutoComplete="false".

alrod:
> @alrod when autocomplete is enabled for single mode instead of batch does the exception get thrown if the lock is lost before autocomplete? or does the run loop swallow this exception?

@elksson for single mode the exception also get thrown if the lock is lost before autocomplete.

elksson:
@alrod I guess that makes sense maybe would could change this to a feature request instead of a bug fix and see how the community feels? 

I would propose that we make a 3rd mode for autocomplete, true, false, and hybrid

When the binding is in "hybrid" mode then the binding would track any calls to release, defer, complete the locks and then skip them when the autocomplete was performed.

alrod:
[...]

PR #55, opened 2020-01-23
Fixed issue that caused the listener loop to throw an exception when …

elksson:
fixes issue #54


alrod:
Closing as discussed in #54

Issue #44, opened 2020-01-05
Blob binding expression to servicebus trigger property no longer works


mliesmons:
In 4.0.0 below worked fine, in 4.1.0, I get a run-time error telling me {BlobContainerAndPath} could not be bound to (this is a public property of the ServiceBusEvent type):

            [ServiceBusTrigger("%ServiceBus:Topic%", "%ServiceBus:Subscription%", Connection = "ConnectionStrings:ServiceBus")] ServiceBusEvent serviceBusEvent,
            [Blob("{BlobContainerAndPath}", FileAccess.Read, Connection = "ConnectionStrings:Blob")] TextReader blobReader)

solonrice:
I also just found that binding expressions fail in Singleton scopes when using a POCO object in the ServiceBusTrigger decorated parameter. Works in 3.1.1 (the version I fell back to).

mliesmons:
> I also just found that binding expressions fail in Singleton scopes when using a POCO object in the ServiceBusTrigger decorated parameter. Works in 3.1.1 (the version I fell back to).

It is awfully quite here, I think we better repost in https://github.com/Azure/azure-webjobs-sdk/issues

PR #62, opened 2020-02-01
Fixing ServiceBusTrigger binding data for POCOs

mathewc:
Fix for Azure/azure-functions-servicebus-extension#44 in collaboration with @alrod.

Relies on the SDK fix https://github.com/Azure/azure-webjobs-sdk/pull/2423. Will need to pull in an updated SDK version with that fix as part of this commit.

### DIFF
--- a/test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
+++ b/test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Runtime.Serialization;
 using System.Text;
 using System.Text.RegularExpressions;
@@ -43,6 +44,7 @@ namespace Microsoft.Azure.WebJobs.Host.EndToEndTests
 
         private static EventWaitHandle _topicSubscriptionCalled1;
         private static EventWaitHandle _topicSubscriptionCalled2;
+        private static EventWaitHandle _eventWait;
 
         // These two variables will be checked at the end of the test
         private static string _resultMessage1;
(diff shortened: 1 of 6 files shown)


test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
```diff
@@ -225,19 +260,25 @@ private async Task<int> CleanUpEntity(string queueName, string connectionString
...
-            await CleanUpEntity(FirstQueueName);
-            await CleanUpEntity(SecondQueueName);
-            await CleanUpEntity(BinderQueueName);
-            await CleanUpEntity(FirstQueueName, _secondaryConnectionString);
-
-            await CleanUpEntity(EntityNameHelper.FormatSubscriptionPath(TopicName, TopicSubscriptionName1));
-            await CleanUpEntity(EntityNameHelper.FormatSubscriptionPath(TopicName, TopicSubscriptionName2));
+            var tasks = new List<Task>
+            {
+                CleanUpEntity(FirstQueueName),
```
mathewc: These tests take way too long to run. At least doing these in parallel will help. We should really refactor most of these tests to be direct invoke (JobHost.CallAsync), since that allows us to test binding permutations quickly. We still need some E2E tests though that actually pull messages, but those can be much fewer.

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
```diff
@@ -626,6 +667,30 @@ public class ServiceBusMultipleMessagesTestJob_BindToPocoArray
             }
         }
 
+        public class ServiceBusArgumentBindingJob
+        {
+            public static void BindToPoco(
+                [ServiceBusTrigger(FirstQueueName)] TestPoco input,
+                string name, string value, string messageId,
```
mathewc: Mix of both POCO and base Message binding data to ensure both work (merged)

alrod:
Also new tests are failed

Issue #64, opened 2020-02-10
Logging shows invalid session id


Zenuka:
In application insights I'm getting strange session ids with some of my trace telemetry. I think it's because of this line: https://github.com/Azure/azure-functions-servicebus-extension/blob/4c4ef3252112bc822a5c80a26824b12146330650/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs#L88

An example session id is:

prop__SessionId | 12/31/9999 11:59:59 PM
-- | --

alrod:
closed

PR #75, opened 2020-04-01
Fix sessionId log

alrod:


alrod:
Fixes https://github.com/Azure/azure-functions-servicebus-extension/issues/64

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerInput.cs
```diff
@@ -83,9 +83,8 @@ public TriggeredFunctionData GetTriggerFunctionData()
                     int[] deliveryCounts = new int[length];
                     DateTime[] enqueuedTimes = new DateTime[length];
                     DateTime[] lockedUntils = new DateTime[length];
-                    string sessionId = string.Empty;
+                    string sessionId = Messages[0].SessionId;
```
soninaren, line 86: Would be good to add some tests.

PR #84, opened 2020-05-18
Add logs for batch sessions

alrod:


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -255,10 +255,13 @@ internal void StartMessageBatchReceiver(CancellationToken cancellationToken)
                         {
                             try
                             {
+                                _logger.LogDebug($"Session accepting..");
                                 receiver = await sessionClient.AcceptMessageSessionAsync();
+                                LogMessageReceiver(receiver, "Session accepted");
```
ConnorMcMahon, line 260: Do we have session id at this point?

alrod: yes, https://github.com/Azure/azure-functions-servicebus-extension/pull/84/files#diff-b5eb5657e38dd115693645d4a26efeceR358

ConnorMcMahon: My bad, should have looked closer.

Issue #85, opened 2020-05-20
Improve Trigger Details


brettsam:
Two things:
1. Today the Scale Controller logs the `EnqueuedSequenceNumber` rather than the `MessageId` -- I think we need to add this to our trigger details so we can better correlate.
2. The timestamp we log should be written with `.ToUniversalTime().ToString("o")` so we have the full details and it's easier to correlate as well.

PR #89, opened 2020-05-29
Reuse IMessageSession and set PrefetchCount for Sessions. Improve Tri…

### LINKED ISSUE #88: Reuse IMessageSession and set PrefetchCount for Sessions

alrod:
…gger Details. Fixes #88, #85.

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -287,27 +287,21 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                                 }
                                 else
                                 {
-                                    // Delivery count is not incremented if 
-                                    // Session is accepted, the messages within the session are not completed (even if they are locked), and the session is closed
-                                    // https://docs.microsoft.com/en-us/azure/service-bus-messaging/message-sessions#impact-of-delivery-count
-                                    if (_isSessionsEnabled)
+                                    List<Task> abandonTasks = new List<Task>();
+                                    foreach (var lockTocken in messagesArray.Select(x => x.SystemProperties.LockToken))
                                     {
-                                        List<Task> abandonTasks = new List<Task>();
-                                        foreach (var lockTocken in messagesArray.Select(x => x.SystemProperties.LockToken))
-                                        {
-                                            abandonTasks.Add(receiver.AbandonAsync(lockTocken));
-                                        }
-                                        await Task.WhenAll(abandonTasks);
+                                        abandonTasks.Add(receiver.AbandonAsync(lockTocken));
                                     }
+                                    await Task.WhenAll(abandonTasks);
                                 }
                             }
-                            else
+                        }
+                        else
+                        {
+                            // Close the session and release the session lock after draining all messages for the accepted session.
+                            if (_isSessionsEnabled)
                             {
-                                // Close the session and release the session lock
-                                if (_isSessionsEnabled)
-                                {
-                                    await receiver.CloseAsync();
-                                }
+                                await receiver.CloseAsync();
                             }
                         }
                     }
@@ -318,7 +312,21 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                     catch (Exception ex)
                     {
                         // Log another exception
-                        _logger.LogError($"An unhandled exception occurred in the message batch receive loop: {ex.ToString()}");
+                        _logger.LogError($"An unhandled exception occurred in the message batch receive loop", ex);
+
+                        if (_isSessionsEnabled && receiver != null)
+                        {
+                            // Attempt to close the session and release session lock to accept a new session on the next loop iteration
+                            try
+                            {
+                                await receiver.CloseAsync();
+                            }
+                            catch
+                            {
+                                // Best effort
+                                receiver = null;
+                            }
+                        }
                     }
                 }
             }, cancellationToken);
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -19,8 +19,6 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
 {
     internal sealed class ServiceBusListener : IListener, IScaleMonitorProvider
     {
-        private const int NumberOfMessagesToReceiveInBatch = 1000;
-
         private readonly MessagingProvider _messagingProvider;
         private readonly ITriggeredFunctionExecutor _triggerExecutor;
         private readonly string _functionId;
@@ -251,11 +249,12 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                             return;
                         }
 
-                        if (_isSessionsEnabled)
+                        if (_isSessionsEnabled && ( receiver == null || receiver.IsClosedOrClosing))
                         {
                             try
                             {
                                 receiver = await sessionClient.AcceptMessageSessionAsync();
+                                receiver.PrefetchCount = _serviceBusOptions.PrefetchCount;
                             }
                             catch (ServiceBusTimeoutException)
                             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -251,11 +249,12 @@ internal void StartMessageBatchReceiver(CancellationToken cancellationToken)
...
-                        if (_isSessionsEnabled)
+                        if (_isSessionsEnabled && ( receiver == null || receiver.IsClosedOrClosing))
                         {
                             try
                             {
                                 receiver = await sessionClient.AcceptMessageSessionAsync();
+                                receiver.PrefetchCount = _serviceBusOptions.PrefetchCount;
                             }
                             catch (ServiceBusTimeoutException)
                             {
```
sidkri, line 260: Would it be useful to have a debug statement here when we there are not sessions to accept or too much of a perf. hit?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -302,11 +301,13 @@ internal void StartMessageBatchReceiver(CancellationToken cancellationToken)
                                     }
                                 }
                             }
-
-                            // Close the session and release the session lock
-                            if (_isSessionsEnabled)
+                            else
                             {
-                                await receiver.CloseAsync();
+                                // Close the session and release the session lock
```
sidkri, line 306: Would be good to enhance this comment and explain that we are only closing the session after draining all prefetched messages.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -302,11 +301,13 @@ internal void StartMessageBatchReceiver(CancellationToken cancellationToken)
                                     }
                                 }
                             }
-
-                            // Close the session and release the session lock
-                            if (_isSessionsEnabled)
+                            else
```
nemakam, line 304: >else [](start = 28, length = 4)

Should this else come after another `}`? Or I might be counting wrong. I assume the intention is to close the receiver only when `messages == null`

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -251,11 +249,12 @@ internal void StartMessageBatchReceiver(CancellationToken cancellationToken)
                             return;
                         }
 
-                        if (_isSessionsEnabled)
+                        if (_isSessionsEnabled && ( receiver == null || receiver.IsClosedOrClosing))
                         {
                             try
                             {
                                 receiver = await sessionClient.AcceptMessageSessionAsync();
+                                receiver.PrefetchCount = _serviceBusOptions.PrefetchCount;
```
nemakam, line 257: There are some rare cases where this returns null. It was a bug long time ago, not sure if its fixed or not. For the sake of precautions, I would put a null check after this and if this happens to be null, do a `continue`

nemakam:
>                                     if (_isSessionsEnabled)

Not related to you PR, but why are we not abandoning the messages in case of non-sessions?

---
Refers to: src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs:293 in 9799ed1. [](commit_id = 9799ed1b7355c4fd027ff59690932c6d62a147b3, deletion_comment = False)

nemakam:
>                                 }

If this throws, you will never end up closing and recreating the session. Lets say you enter into SessionLockLost exception once, and then you'd never recovver from that

---
Refers to: src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs:287 in 9799ed1. [](commit_id = 9799ed1b7355c4fd027ff59690932c6d62a147b3, deletion_comment = False)

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
sidkri: Good catch, I just counted and it does need to be one level of nesting out (after line 312)

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: I think there can be too much noise in the logs

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: Comment is enhanced

alrod:
> > ```
> >                                 if (_isSessionsEnabled)
> > ```
> 
> Not related to you PR, but why are we not abandoning the messages in case of non-sessions?
> 
> Refers to: src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs:293 in [9799ed1](https://github.com/Azure/azure-functions-servicebus-extension/commit/9799ed1b7355c4fd027ff59690932c6d62a147b3). [](commit_id = [9799ed1](https://github.com/Azure/azure-functions-servicebus-extension/commit/9799ed1b7355c4fd027ff59690932c6d62a147b3), deletion_comment = False)

We are relying on lock token expiration in case of non-session. This code was added because delivery count is not increased in case of sessions some time. I think it's also good to manually abandon non-session messages. So I remove `if (_isSessionsEnabled)` as part of this PR.

alrod:
> > ```
> >                             }
> > ```
> 
> If this throws, you will never end up closing and recreating the session. Lets say you enter into SessionLockLost exception once, and then you'd never recovver from that
> 
> Refers to: src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs:287 in [9799ed1](https://github.com/Azure/azure-functions-servicebus-extension/commit/9799ed1b7355c4fd027ff59690932c6d62a147b3). [](commit_id = [9799ed1](https://github.com/Azure/azure-functions-servicebus-extension/commit/9799ed1b7355c4fd027ff59690932c6d62a147b3), deletion_comment = False)

From  https://docs.microsoft.com/en-au/azure/service-bus-messaging/message-sessions#session-features:
_The lock is released when Close or CloseAsync are called, or when the lock expires in cases in which the application is unable to do the close operation._
When will the session lock expire? Will IMessageSession close automatically on session lock expire?

I added code to close IMessageSession in catch block.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
alrod: There will be a NullReferenceException and we will try to accept a new session on the next loop iteration.

sidkri:
Since the default PrefetchCount will remain 0 after this change, I recommend we lower the default value of "MaxMessageCount" to something more likely to be streamed in within the 20ms window. This of course depends on message size but defaulting to a lower value will work for most users and users that want a higher batch size can change the value and when we document that PrefetchCount should be set to the same or higher value, they will hopefully see that avoid running into issues with no prefetch and large batch sizes.  Based on testing so far, a default value around 5 or 10 sounds best.  When we provide a Batch specific PrefetchCount configuration option and default it the same default as MaxMessageCount, we can increase the default values to a higher value.

alrod:
> Since the default PrefetchCount will remain 0 after this change, I recommend we lower the default value of "MaxMessageCount" to something more likely to be streamed in within the 20ms window. This of course depends on message size but defaulting to a lower value will work for most users and users that want a higher batch size can change the value and when we document that PrefetchCount should be set to the same or higher value, they will hopefully see that avoid running into issues with no prefetch and large batch sizes. Based on testing so far, a default value around 5 or 10 sounds best. When we provide a Batch specific PrefetchCount configuration option and default it the same default as MaxMessageCount, we can increase the default values to a higher value.

As discussed we a re not going to change default values.

nemakam:
@alrod, I took a look at the code again, and it looks good to me as well. (y0

Issue #97, opened 2020-06-30
[ServiceBusListener.StartMessageBatchReceiver] Exception details not getting logged


tristan-earl:
We're seeing the following error in our Application Insights logs:
`"An unhandled exception occurred in the message batch receive loop"`.

I found [this line in ServiceBusListener.cs](https://github.com/Azure/azure-functions-servicebus-extension/blob/8a01c68d16c5eeeda80010fd32f34d265eca9534/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs#L315), which logs the message along with the exception.  However, the exception details are not showing up in the logs.  I notice that this overload of `LogError` is being called:

`void LogError(this ILogger logger, string message, params object[] args)` 

...instead of this one:

`void LogError(this ILogger logger, Exception exception, string message, params object[] args)`

Since the Exception isn't explicitly passed in, I suspect it's just getting ignored.  I know when the latter overload is called, App Insights will track it as an exception.  Can this be changed so we get the full exception details in our logs?

Zenuka:
Should be a simple fix, right?

PR #100, opened 2020-08-06
Log exception

Zenuka:
Include exception when logging an error in the ServiceBusListener

Fixes: Azure/azure-functions-servicebus-extension#97


Zenuka:
My first PR, hope I did everything right 🤞

Zenuka:
Can this one-line change be added to the next release @sidkri or @alrod? Too bad it didn't make the release 6 days ago. Fixes #97 and #101.

Zenuka:
Just ran into this again today. Something was dead-lettering a lot of our service bus messages but no exceptions or alerts where fired because the exception is not logged. The only thing that pops-up in application insights is a single trace message with the LogLevel set to `Error` like so:
image

Can we get this released so that the exception is logged, @alrod?

alrod (approved):

PR #102, opened 2020-08-11
Graceful shutdown (message draining)

sidkri:
Azure functions will support draining in-flight messages before shutdown of the runtime.  The current extension implementation closes the connection to Service Bus during shutdown before allowing executing function invocations to complete processing and prevents these messages from being completed or abandoned.  Additionally, new messages start processing up until the connection to Service Bus is closed.

This PR implements graceful shutdown in the extension by leveraging a new set of "Unregister" methods in the Service Bus SDK : https://github.com/Azure/azure-sdk-for-net/pull/14021 and published in 4.2.0 and 5.0.0 of the Service Bus SDK used by this library.

This extension supports single message as well as batch processing and the message draining implementation is different for each path.

**Batch processing:**
The ServiceBusListener when created in batch mode, polls Service Bus for messages in a loop while checking to ensure that it is still in started state and no cancellation has been requested.  The new Functions runtime "/drain" API will be invoked when the  runtime needs to shutdown and sufficient time will be given for any functions that have already been invoked to complete.  The invocation of "/drain" calls StopAsync() on all listeners and with this change, the connections to Service Bus will remain open until the runtime is shutdown and Dispose() is called.

**Single message processing:**
The new QueueClient.UnregisterSessionHandlerAsync, SubscriptionClient.UnregisterSessionHandlerAsync and MessageReceiver.UnregisterMessageHandlerAsync methods are called when the extension is stopped.  These calls block until all in-flight messages complete processing (drain) and new messages no longer get dispatched.  As the amount of time given to functions to execute depends on the SKU, a MaxValue timespan is specified when draining as the host will shut down regardless after the max function execution time.

Other changes:
1. Addressed #95 
2. Support for logging to test output via ITestOutputHelper in both Integration test classes.
3. All test settings are exported to environment variables as WebJobs relies solely on environment variables.  This obviates the need to set environment variables at the operating system level when debugging tests locally in Visual Studio.
4. Stopping and disposing the host object in each test consistently.


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -38,6 +38,7 @@ internal sealed class ServiceBusListener : IListener, IScaleMonitorProvider
         private ClientEntity _clientEntity;
         private bool _disposed;
         private bool _started;
+        private bool _isStopping;
```
alrod, line 41: can we use one flag for single and batch dispatch?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
```diff
@@ -6,14 +6,15 @@
     <RootNamespace>Microsoft.Azure.WebJobs.ServiceBus</RootNamespace>
     <PackageId>Microsoft.Azure.WebJobs.Extensions.ServiceBus</PackageId>
     <Description>Microsoft Azure WebJobs SDK ServiceBus Extension</Description>
-    <Version>4.1.2</Version>
+    <Version>4.2.0$(VersionSuffix)</Version>
```
alrod, line 9: where is `VersionSuffix `defined?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
```diff
@@ -6,14 +6,15 @@
...
     <CommitHash Condition="$(CommitHash) == ''">N/A</CommitHash>
     <InformationalVersion>$(Version) Commit hash: $(CommitHash)</InformationalVersion>
     <Authors>Microsoft</Authors>
     <Company>Microsoft</Company>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageLicenseUrl>https://go.microsoft.com/fwlink/?linkid=2028464</PackageLicenseUrl>
     <PackageIconUrl>https://raw.githubusercontent.com/Azure/azure-webjobs-sdk/dev/webjobs.png</PackageIconUrl>
+    <PackageIcon>webjobs.png</PackageIcon>
```
alrod, line 17: why do we need the image?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
```diff
@@ -28,6 +29,7 @@
     <StyleCopTreatErrorsAsWarnings>false</StyleCopTreatErrorsAsWarnings>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
     <WarningsAsErrors />
+    <NoWarn>1701;1702;NU5125;NU5048</NoWarn>
```
alrod, line 32: why the warnings were excluded?

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -124,20 +125,45 @@ public async Task StopAsync(CancellationToken cancellationToken)
...
-                _clientEntity = null;
+                _isStopping = true;
+
+                if (_isSessionsEnabled)
+                {
+                    if (_clientEntity != null)
+                    {
+                        if (_clientEntity is QueueClient queueClient)
+                        {
+                            await queueClient.UnregisterSessionHandlerAsync(TimeSpan.MaxValue);
```
alrod, line 147: what does` TimeSpan.MaxValue` parameter mean?

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
```diff
@@ -212,6 +228,65 @@ public async Task BindToString()
...
+            _drainValidationPreDelay = new ManualResetEvent(initialState: false);
+            _drainValidationPostDelay = new ManualResetEvent(initialState: false);
+
+            if (sendToQueue)
+            {
+                await WriteQueueMessage(_primaryConnectionString, FirstQueueName, "queue-message-draining-no-sessions-1");
+            }
+            else
+            {
+                await WriteTopicMessage(_primaryConnectionString, TopicName, "topic-message-draining-no-sessions-1");
```
alrod, line 273: add "topic-message-draining-no-sessions-1" to constants

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusSessionsEndToEndTests.cs
```diff
@@ -344,13 +349,78 @@ public async Task TestBatch_DataContractPoco()
...
+                await ServiceBusEndToEndTests.WriteTopicMessage(
+                    _connectionString, _topicName, "topic-message-draining-with-sessions-1", _drainModeSessionId);
+            }
+
+            // Wait to ensure function invocatoin has started before draining messages
+            Assert.True(_drainValidationPreDelay.WaitOne(SBTimeout));
+
+            // Start draining in-flight messages
+            var drainModeManager = host.Services.GetService<IDrainModeManager>();
+            await drainModeManager.EnableDrainModeAsync(CancellationToken.None);
```
alrod, line 404: do we call `StopAsync` for the SB listener here?

alrod (changes requested):

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
sidkri: This flag is different from _started, it's used to prevent calling of StopAsync while a stop is in progress.  In batch mode, the operation is quick and its ok to get called multiple times but in single dispatch mode, we should not call Unregister* while its in progress.  I can alternately add a lock and check _started flag which is a better way to handle concurrency.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
sidkri: This is normally blank and is used when debugging locally.  WebJobs common.props uses this mechanism as well.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
sidkri: This is to fix the warning: https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5048

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
sidkri: This is to suppress the warning: https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu5125 until we change the way we link to the license.  I can remove the suppression if we don't want it in the repo.

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusSessionsEndToEndTests.cs
sidkri: Yes.

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/ServiceBusEndToEndTests.cs
sidkri: Done.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
sidkri: This is how long UnregisterSessionHandlerAsync will allow messages to drain (finish processing) before timing out the operation.  As the amount of time functions host will allow functions to complete processing before shutting down depends on the SKU, this ensures we don't prematurely terminate in-flight messages without having to dynamically set a timeout value depending on the SKU (which is unknown by the extension code).  If triggered functions exceed the SLA for the SKU, the host will shutdown anyway which closes the Service Bus connection.

soninaren (approved):
Just a nit comment. Looks good otherwise.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
sidkri: Resolved by resolving merge conflicts.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/WebJobs.Extensions.ServiceBus.csproj
sidkri: Reverted

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/README.md
```diff
@@ -0,0 +1,34 @@
+﻿# Service Bus Extension for Azure Functions guide to running integration tests locally
+Integration tests are implemented in the `EndToEndTests` and `SessionsEndToEndTests` classes and require special configuration to execute locally in Visual Studio or via dotnet test.  
+
+All configuration is done via a json file called `appsettings.tests` which on windows should be located in the `%USERPROFILE%\.azurefunctions` folder (e.g. `C:\Users\user123\.azurefunctions`)
+
+**Note:** *The specifics of the configuration will change when the validation code is modified so check the code for the latest configuration if the tests do not pass as this readme file may not have been updated with each code change.*
+
+Create the appropriate Azure resources if needed as explained below and create or update the `appsettings.tests` file in the locatoin specified above by copying the configuration below and replacing all the `PLACEHOLDER` values
```
mhoeger, line 8: nit: ~locatoin~ location

sidkri: Thanks, fixed.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -118,27 +120,53 @@ public Task StartAsync(CancellationToken cancellationToken)
...
+                    // as the amount of time Service Bus SDK should wait for in-flight messages to complete procesing after 
+                    // unregistering the message handler so that functions have as long as the host continues to run time to complete.
+                    if (_singleDispatch)
+                    {
+                        if (_isSessionsEnabled)
+                        {
+                            if (_clientEntity != null)
+                            {
+                                if (_clientEntity is QueueClient queueClient)
+                                {
```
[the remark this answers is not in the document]
sidkri: Did you mean lines 140 and 142?  If so, I don't want lines 148 and 149 executed on a null _clientEntity so can't check in a single if statement.  Would definitely like to compact this if chain!

test/Microsoft.Azure.WebJobs.Extensions.ServiceBus.Tests/README.md
```diff
@@ -0,0 +1,34 @@
+﻿# Service Bus Extension for Azure Functions guide to running integration tests locally
```
alrod, line 1: Nice!

alrod (approved):

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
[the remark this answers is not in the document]
soninaren: I see. You probably can make it easier to read by checking for negation instead. Something like.
```
if (!_singleDispatch)
{
 return:
}
```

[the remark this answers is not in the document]
sidkri: have to set the _started flag to false in all cases.  Would have to duplicate that logic in two places with the short circuit of (!_singleDispatch) so not sure if that's a better option as more common steps may be added at a later time and those would have to be duplicated.

PR #111, opened 2020-09-22
Convert ExpiresAtUtc to UTC time to mitigate serialization issue

amamounelsayed:
This issue it related to Java and Python can't parse the date for ExpiresAtUtc. 
Python issue details https://github.com/Azure/azure-functions-python-worker/issues/460

We believe the issue is coming from this line [Message.cs:183](https://github.com/Azure/azure-sdk-for-net/blob/f83d1822f337c7eb2d8d4014b4932eb81a9b2979/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Message.cs#L183)
We can see this line does not return time in UTC like line 186.

This is temporary until the Service Bus SDK addresses the missing timezone issue in case DateTime.MaxValue
https://github.com/Azure/azure-sdk-for-net/issues/15343

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerBindingStrategy.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerBindingStrategy.cs
@@ -125,7 +125,8 @@ namespace Microsoft.Azure.WebJobs.ServiceBus
                 deliveryCounts[i] = messages[i].SystemProperties.DeliveryCount;
                 deadLetterSources[i] = messages[i].SystemProperties.DeadLetterSource;
                 lockTokens[i] = messages[i].SystemProperties.IsLockTokenSet ? messages[i].SystemProperties.LockToken : string.Empty;
-                expiresAtUtcs[i] = messages[i].ExpiresAtUtc;
+                //this is temporary until the Service Bus SDK addresses the missing timezone issue in case DateTime.MaxValue, github.com/Azure/azure-sdk-for-net/issues/15343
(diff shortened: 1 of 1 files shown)


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Triggers/ServiceBusTriggerBindingStrategy.cs
```diff
@@ -125,7 +125,7 @@ internal static void AddBindingData(Dictionary<string, object> bindingData, Mess
                 deliveryCounts[i] = messages[i].SystemProperties.DeliveryCount;
                 deadLetterSources[i] = messages[i].SystemProperties.DeadLetterSource;
                 lockTokens[i] = messages[i].SystemProperties.IsLockTokenSet ? messages[i].SystemProperties.LockToken : string.Empty;
-                expiresAtUtcs[i] = messages[i].ExpiresAtUtc;
+                expiresAtUtcs[i] = messages[i].ExpiresAtUtc.ToUniversalTime();
```
sidkri, line 128: Please leave a comment indicating this is temporary until the Service Bus SDK addresses the missing timezone issue.  Same for the other area in this file.

amamounelsayed: Done

amamounelsayed:
Tested scenarios, java function has unlimited timetolive for batch and single message.
Also 14 days timetolive for batch and single message.

sidkri (approved):

alrod (approved):

amamounelsayed:
Fixes https://github.com/Azure/azure-functions-java-worker/issues/403

PR #121, opened 2020-11-09
Fix PeekAsync for ServiceBusScaleMonitor

wenhzha:
As suggested by Service Bus team in the service bus runtime driven scaling investigation ([ICM ticket](https://portal.microsofticm.com/imp/v3/incidents/details/202193612/home)),

 ```message = await _receiver.Value.PeekAsync()```

 intends to peek the first message . However, in this usage, the client remembera the sequence number of the last message and peeks the next message on the next call. For example, if there are two messages in the queue, calling peekAsync returns the first message. If you call peekAsync again, it returns the second message. If you call peekAsync again, it will return no message or null. 

Use PeekBySequenceNumberAsync instead so it forces the client to check the first message from sequence number =0. 

Repro:
1. drop two message on the service bus queue/topic
2. reuse the service bus message receiver and call peek async three times

Repro-ed and verified the PeekBySequenceNumberAsync would switch to the desired behavior. 

### Pull request checklist

    * [ ] Otherwise: Documentation issue linked to PR. Update [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=csharp) to reflect new configuration options
* [x] My changes **should not** be added to the release notes for the next release
    * [ ] Otherwise: I've added my notes to `release_notes.md`
* [ ] I have added all required tests (Unit tests, E2E tests)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusScaleMonitor.cs
@@ -66,7 +66,9 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             try
             {
                 // Peek the first message in the queue without removing it from the queue
-                message = await _receiver.Value.PeekAsync();
+                // PeekAsync remembers the sequence number of the last message, so the second call returns the second message instead of the first one
+                // Use PeekBySequenceNumberAsync with fromSequenceNumber = 0 to always get the first available message
+                message = await _receiver.Value.PeekBySequenceNumberAsync(0);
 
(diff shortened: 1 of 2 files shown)


gzuber (approved):

sidkri (changes requested):
Checking to see how this was verified.  Sequence numbers embed partition information and are increasing within a partition.  I could not find special treatment for a zero sequence number in the ServiceBus source code.

sidkri (approved):
Discussed offline to understand how this change was validated. @wenhzha please update the PR with details of how you validated the change since we don't have existing integration test coverage for the scale monitor logic.

alrod (approved):

mathewc:
Thanks for the investigation and fix @wenhzha!

PR #143, opened 2021-03-30
Allow all in-flight Service Bus messages to complet during shutdown when function execution completes

sidkri:
Prior to the changes in this PR, when a WebJob starts stopping, Service Bus messages continue to start processing but get abandoned as ServiceBusListener.Cancel() is called first which cancels the cancellation token source related to function invocation but does not stop the underlying Service Bus SDK from processing new messages. Additionally, the default MessageProcessor implementation throws an exception if the cancellation token involved in function execution is set which results in all in-flight messages getting abandoned even though they were successfully processed by the function.  Both behaviors are modified as explained in the changes below.

Changes:
- Modified behavior of ServiceBusListener.Cancel() to call StopAsync() instead of cancelling the cancellation token as this was inconsistent with other extensions and impacting WebJobs on shutdown.
- When the cancellation token passed to functions is cancelled, messages are no longer automatically abandoned by the default MessageProcessor implementation.  This behavior was changed to allow in-flight function executions to complete successfully via "drain mode" which will provide functions with an extended period of time when an instance is shutting down as a result of scaling in.

### Pull request checklist
    * [ ] Otherwise: Documentation issue linked to PR. Update [documentation](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-service-bus-trigger?tabs=csharp) to reflect new configuration options
* [] My changes **should not** be added to the release notes for the next release
    * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] I have added all required tests (Unit tests, E2E tests) - No new tests needed

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -2,9 +2,9 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR)
 -->
-#### Version 4.2.1
--  Upgraded to version 4.2.1 of Microsoft.Azure.ServiceBus dependency (PR# 133)
--  Bug fix in scaling related metrics gathering logic (PR# 121)
+#### Version 4.2.2
+- Modified behavior of ServiceBusListener.Cancel() to call StopAsync() instead of cancelling the cancellation token as this was inconsistent with other extensions and impacting WebJobs on shutdown.
+- When the cancellation token passed to functions is cancelled, messages are no longer automatically abandoned by the default MessageProcessor implementation.  This behavior was changed to allow in-flight function executions to complete successfully via "drain mode" which will provide functions with an extended period of time when an instance is shutting down as a result of scaling in.
 
-**Release sprint:** Sprint 94
-[ [bugs](https://github.com/Azure/azure-functions-servicebus-extension/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+94%22+label%3Abug+is%3Aclosed) | [features](https://github.com/Azure/azure-functions-servicebus-extension/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+94%22+label%3Afeature+is%3Aclosed) ]
+**Release sprint:** Sprint 98
+[ [bugs](https://github.com/Azure/azure-functions-servicebus-extension/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+98%22+label%3Abug+is%3Aclosed) | [features](https://github.com/Azure/azure-functions-servicebus-extension/issues?q=is%3Aissue+milestone%3A%22Functions+Sprint+98%22+label%3Afeature+is%3Aclosed) ]
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -170,7 +170,6 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
         public void Cancel()
         {
             ThrowIfDisposed();
-            StopAsync(CancellationToken.None).Wait();
         }
 
         [SuppressMessage("Microsoft.Usage", "CA2213:DisposableFieldsShouldBeDisposed", MessageId = "_cancellationTokenSource")]
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -124,41 +124,39 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             {
                 try
                 {
-                    if (!_started)
+                    if (_started)
                     {
-                        throw new InvalidOperationException("The listener has not yet been started or has already been stopped.");
-                    }
-
-                    // Unregister* methods stop new messages from being processed while allowing in-flight messages to complete.
-                    // As the amount of time functions are allowed to complete processing varies by SKU, we specify max timespan
-                    // as the amount of time Service Bus SDK should wait for in-flight messages to complete procesing after 
-                    // unregistering the message handler so that functions have as long as the host continues to run time to complete.
-                    if (_singleDispatch)
-                    {
-                        if (_isSessionsEnabled)
+                        // Unregister* methods stop new messages from being processed while allowing in-flight messages to complete.
+                        // As the amount of time functions are allowed to complete processing varies by SKU, we specify max timespan
+                        // as the amount of time Service Bus SDK should wait for in-flight messages to complete procesing after 
+                        // unregistering the message handler so that functions have as long as the host continues to run time to complete.
+                        if (_singleDispatch)
                         {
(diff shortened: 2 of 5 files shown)


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -172,7 +170,7 @@ public async Task StopAsync(CancellationToken cancellationToken)
         public void Cancel()
         {
             ThrowIfDisposed();
-            _cancellationTokenSource.Cancel();
+            StopAsync(CancellationToken.None).Wait();
```
alrod, line 173: I think we have the `CancellationToken` for graceful shutdown:
https://docs.microsoft.com/en-us/azure/azure-functions/functions-dotnet-class-library?tabs=v2%2Ccmd#cancellation-tokens

alrod:
How does the job host stop the listener? Does it call `Cancel` and then `StopAsync`?
From the interface description we are just should cancel "in process" operations https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Listeners/IListener.cs#L32

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/MessageProcessor.cs
```diff
@@ -71,8 +71,6 @@ public virtual Task CompleteProcessingMessageAsync(Message message, FunctionResu
                 throw new ArgumentNullException(nameof(result));
             }
 
-            cancellationToken.ThrowIfCancellationRequested();
```
alrod, line 74: we have the same pattern in `SessionMessageProcessor`, do we want to change it also?

sidkri:
> How does the job host stop the listener? Does it call `Cancel` and then `StopAsync`?
> From the interface description it should just cancel "in process" operations https://github.com/Azure/azure-webjobs-sdk/blob/dev/src/Microsoft.Azure.WebJobs.Host/Listeners/IListener.cs#L32

Per the interface description "Cancel any in progress listen operation" we should just cancel in-flight executions but listening will continue and it's not clear what should happen with new messages. The issue in this extension is that after Cancel() is called even new messages that start processing after the call will have a cancelled token.  This is appropriate from the perspective of notifying the function being executed that execution may get interrupted/cancelled at any time after Cancel() is invoked.  The problem is that the Service Bus extension [checks the cancellation token while binding](https://github.com/Azure/azure-functions-servicebus-extension/blob/32f8df8a62c38cb5ae4333875d73d60f981045a3/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Bindings/ServiceBusBinding.cs#L42) as well as [when execution completes](https://github.com/Azure/azure-functions-servicebus-extension/blob/32f8df8a62c38cb5ae4333875d73d60f981045a3/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/MessageProcessor.cs#L74) and in both cases throws an exception if the cancellation token is set, presumably so that the message is abandoned instead of completed as execution may get interrupted (ServiceBusBinding) or might have been skipped in the function (MessageProcessor).  Another pattern we can follow is to simply notify the function code via the cancellation token that execution may get cancelled and let it decide to abandon the service bus message by throwing an exception (when autoComplete is enabled).  Remove the check and throw logic in the two places I linked above could be a better change.  I opted for modifying the behavior of ServiceBusListener:Cancel() to align with that of the EventHubsListener.Cancel() which simply calls StopAsync().

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/MessageProcessor.cs
sidkri: Good catch!   Updated that class as well.

sidkri: FYI, I have removed this logic as it impacts drain mode by abandoning all in-flight messages (execution started before but did not complete by the time StopAsync() is called) which defeats the purpose of the draining logic.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -124,41 +124,39 @@ public async Task StopAsync(CancellationToken cancellationToken)
             {
                 try
                 {
-                    if (!_started)
+                    if (_started)
```
alrod, line 127: How did it work before? Did we never stop the started listener?

sidkri: There is a flag in check in [FunctionListener](https://github.com/Azure/azure-webjobs-sdk/blob/cc7522d2ea1d1c2ac667ab7490fa37460e4af53a/src/Microsoft.Azure.WebJobs.Host/Listeners/FunctionListener.cs#L139) that wraps this listener so from WebJobs/Functions side it will not call this twice (and encounter the exception).  Now we are calling it from Cancel so its important to not throw an exception here.  Its also consistent with other extensions.

alrod:
Adding some conversation for history:

```
[4/8 6:14 PM] Alexey Rodionov
 I thought more about the change

why do we remove cancellationToken.ThrowIfCancellationRequested(); in MessageProcessor and SessionMessageProcessor and leave them in other places?  Maybe we should also leave them so on after dispose with not get disposed object exception on message comlete/abandon?
	If now the token is canceled on dispose, will the customer have the same issues as we still cancelling the token?

​[4/8 6:50 PM] Sid Krishna
    #1 I need to remove them from processors as Drain mode manager cancel's the token
​[4/8 6:51 PM] Sid Krishna
    #2 dispose is called at the end, cancel happens at the start of shutdown
​[4/8 6:51 PM] Sid Krishna
    the tests are manually creating an IHostedService and stopping it
​[4/8 6:51 PM] Sid Krishna
    WebJobs itself has a different lifecycle
​[4/8 6:52 PM] Sid Krishna
    I'll validate #1 again, the token that drain manager cancels is an further down the stack so might not affect message processor actually
​[4/8 7:48 PM] Sid Krishna
    I tested that the cancellation token does not get set when draining at the message processor level (your #1) so those removals are actually not needed so I reverted those changes and pushed.
​[4/8 7:51 PM] Sid Krishna
    That leaves your #2 which I think is not an issue as by the time the object is getting disposed the process is close to stopping so impact will be minimal if at all v/s when Cancel is invoked for which there is a long delay between that and process stopping.
```

alrod (approved):

release_notes.md
```diff
@@ -2,9 +2,9 @@
 <!-- Please add your release notes in the following format:
 - My change description (#PR)
 -->
-#### Version 4.2.1
--  Upgraded to version 4.2.1 of Microsoft.Azure.ServiceBus dependency (PR# 133)
--  Bug fix in scaling related metrics gathering logic (PR# 121)
+#### Version 4.2.2
+- Modified behavior of ServiceBusListener.Cancel() to call StopAsync() instead of cancelling the cancellation token as this was inconsistent with other extensions and impacting WebJobs on shutdown.
```
mathewc, line 6: These notes are out of date?

PR #171, opened 2021-10-12
Checking in a pull request template

sidkri:
Based off azure-functions-host

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,20 @@
+<!-- Please provide all the information below.  -->
+
+### Issue describing the changes in this PR
+
+Resolves #issue_for_this_pr
+
+### Pull request checklist
+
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,20 @@
+<!-- Please provide all the information below.  -->
+
+### Issue describing the changes in this PR
+
+Resolves #issue_for_this_pr
```
liliankasem, line 5: Fabio made the comment that "For GH keywords, most of our templates standardize on using them with no casing variation."

So I think we're leaving this as a lowercase to stay consistent 

```suggestion
resolves #issue_for_this_pr
```

liliankasem (approved):

.github/pull_request_template.md
sidkri: fixed.

PR #184, opened 2022-04-04
Fix batch receive race condition

### LINKED ISSUE #147: Fix Batch receive race condition
A race condition was identified and fixed in the T2 extension https://github.com/Azure/azure-sdk-for-net/pull/20288. We need to fix in T1 as well.

hossam-nasr:
### Issue describing the changes in this PR

Resolves #147 , mirror of track 2 PR: https://github.com/Azure/azure-sdk-for-net/pull/20288

### Pull request checklist

  * [ ] Otherwise: Documentation issue linked to PR
* [x] My changes **should not** be added to the release notes for the next release
  * [ ] Otherwise: I've added my notes to `release_notes.md`
* [x] I have added all required tests (Unit tests, E2E tests)

### DIFF
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -88,6 +88,11 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                 throw new InvalidOperationException("The listener has already been started.");
             }
 
+            // The _started flag is set here and not after StartMessageBatchReceiver is called
+            // to avoid a race condition in which the batch receive loop could start before the _started
+            // flag is set to true, leading to the listener never receiving any messages 
+            _started = true;
+
             if (_singleDispatch)
             {
                 if (_isSessionsEnabled)
@@ -112,7 +117,6 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
             {
                 StartMessageBatchReceiver(_cancellationTokenSource.Token);
             }
-            _started = true;
 
             return Task.CompletedTask;
         }
@@ -155,8 +159,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                                 }
                             }
                         }
-                        // Batch processing will be stopped via the _cancellationTokenSource on its next iteration
-                        _cancellationTokenSource.Cancel();
+                        // Batch processing will be stopped via the _started flag on its next iteration
                     }
 
                     _started = false;
@@ -275,7 +278,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                 {
                     try
                     {
-                        if (cancellationToken.IsCancellationRequested)
+                        if (!_started || cancellationToken.IsCancellationRequested)
                         {
                             _logger.LogInformation("Message processing has been stopped or cancelled");
                             return;
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -155,10 +155,10 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                                 }
                             }
                         }
+                        // Batch processing will be stopped via the _cancellationTokenSource on its next iteration
+                        _cancellationTokenSource.Cancel();
                     }
 
-                    // Batch processing will be stopped via the _cancellationTokenSource on its next iteration
-                    _cancellationTokenSource.Cancel();
                     _started = false;
                 }
                 finally
--- a/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
+++ b/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
@@ -155,9 +155,10 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                                 }
                             }
                         }
-                        // Batch processing will be stopped via the _started flag on its next iteration
                     }
 
+                    // Batch processing will be stopped via the _cancellationTokenSource on its next iteration
+                    _cancellationTokenSource.Cancel();
                     _started = false;
                 }
                 finally
@@ -274,7 +275,7 @@ namespace Microsoft.Azure.WebJobs.ServiceBus.Listeners
                 {
                     try
                     {
-                        if (!_started || cancellationToken.IsCancellationRequested)
+                        if (cancellationToken.IsCancellationRequested)
                         {
                             _logger.LogInformation("Message processing has been stopped or cancelled");
                             return;


src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -274,7 +275,7 @@ internal void StartMessageBatchReceiver(CancellationToken cancellationToken)
                 {
                     try
                     {
-                        if (!_started || cancellationToken.IsCancellationRequested)
+                        if (cancellationToken.IsCancellationRequested)
                         {
                             _logger.LogInformation("Message processing has been stopped or cancelled");
```
sidkri, line 280: The log message still says "has been stopped"

sidkri: Just checked the track 2 code and unfortunately it has the same log message that is not modified to reflect new logic so for consistency we can consider leaving this the same.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
```diff
@@ -155,9 +155,10 @@ public async Task StopAsync(CancellationToken cancellationToken)
                                 }
                             }
                         }
-                        // Batch processing will be stopped via the _started flag on its next iteration
                     }
 
+                    // Batch processing will be stopped via the _cancellationTokenSource on its next iteration
+                    _cancellationTokenSource.Cancel();
```
sidkri, line 161: This ideally should be done after line 157 (within the if(_started) block

sidkri: Also, this is only used to stop batch processing so should be in the else clause of if (_started).  It of course most likely does not have any adverse effects but is not consistent logic wise.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
JoshLove-msft: It is still correct that it was stopped or cancelled, as the cancellation token is signalled when stopping.

src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs
JoshLove-msft: This is cancelled unconditionally in the Dispose method so it is consistent to cancel it unconditionally here I think. Also, this does impact the single dispatch case because the cancellation token used in TryExecuteAsync is linked with this one.

sidkri: Hi @JoshLove-msft , we have a state in between of running and stopped called "draining" and when we initiate draining, we call the StopAsync() on each listener and wait for all invocation to complete before actually stopping the worker.  We allow up to an hour for this depending on SKU.  When draining invocations, we don't want to signal a cancellation to customer code as we will not interrupt execution of their code.  We don't cancel the cancellation token in StopAsync() for single dispatch mode either.  This only happens in DisposeAsync().

I'm recommending the following fix for the race condition issue:

In StartAsync() set the _started flag to true at the end of the if (_singleDispatch) block and in the corresponding else block (batch processing) either set the flag before calling StartMessageBatchReceiver(...) or have this method set the flag before starting the while (true) loop.

JoshLove-msft: But isn't the cancellation token flowed through in the underlying SDK in a cancelled state when unregistering?

JoshLove-msft: Another question for you @sidkri - should we be unregistering from Cancel as well?

sidkri: I'm not following this part: "isn't the cancellation token flowed through in the underlying SDK in a cancelled state when unregistering?".  Could you clarify which token you are referring to through references to the code?  There are quite a few cancellation tokens in the listener class :)

The cancellation token we pass to the user code in ProcessMessageAsync() [_triggerExecutor.TryExecuteAsync(data, linkedCts.Token)] is not cancelled while unregistering.  Unregistering (in track 1 extension) is done in the StopAsync() call and we don't cancel any of the this token sources there.  This behavior is preserved in track 2 as well.  Not sure if you were asking referring to this logic.

I could not find any place that we call ServiceBusListener.Cancel() so I am not sure about its purpose.  I can see how it makes sense to unregister there as all future invocations would have the token cancelled anyway but semantically it doesn't feel right to unregister there.  It's possible to think of a scenario where we continue invoking the customer code with a cancelled token.   Not knowing the purpose of this, I would not change its existing behavior.

JoshLove-msft: > I'm not following this part: "isn't the cancellation token flowed through in the underlying SDK in a cancelled state when unregistering?". Could you clarify which token you are referring to through references to the code? There are quite a few cancellation tokens in the listener class :)

Indeed there are :)
My question is about this linked token - https://github.com/Azure/azure-functions-servicebus-extension/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs#L223

Although the extension is not cancelling the token when stopping, it is my understanding that the token that comes into the registered message callback here - https://github.com/Azure/azure-functions-servicebus-extension/blob/dev/src/Microsoft.Azure.WebJobs.Extensions.ServiceBus/Listeners/ServiceBusListener.cs#L221 would actually be cancelled by the Track 1 SDK. Because of this, I'm not sure what benefit we gain in the extension if we don't cancel the extension token, as the customer is getting the linked token anyways which is already canceled.

sidkri: I see your point now.  My understanding was that this token passed by the Service Bus SDK is not cancelled when the handler is unregistered via MessageReceiver.UnregisterMessageHandlerAsync() - I worked with the service bus team on developing this functionality a couple of years ago for track 1 and had requested that the token not be cancelled as we are allowing the message processing to complete and not interrupting them (graceful shutdown via draining).  Here is the relevant code:

[Draining logic inside UnregisterMessageHandlerAsync()](https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs#L935)

Cancellation of the token source happens after all inflight tasks are done:
https://github.com/Azure/azure-sdk-for-net/blob/4162f6fa2445b2127468b9cfd080f01c9da88eba/sdk/servicebus/Microsoft.Azure.ServiceBus/src/Core/MessageReceiver.cs#L950

runningTaskCancellationTokenSource's token is what is passed to ProcessMessageAsync() (and variants), the "receivePumpCancellationTokenSource" is different.

We can of course verify this via a sample program if needed to be sure but looking at the source code I'm pretty sure about this.

JoshLove-msft: I see - I stand corrected. This means that we have slightly different semantics between T1 and T2 around what a cancelled token means. I should ask though - if the token is only cancelled after all in-flight processing is done, what is the benefit of passing the token to the users message handler and function in the first place?

JoshLove-msft: I guess the idea is that they may want to hold onto the token beyond the scope of their function execution?

sidkri: That is a good question about when the token actually gets cancelled. It will be cancelled in these scenarios:
1. By the extension when calling "Cancel()" on the listener will cancel the token - I'm not sure if this method is actually used in Functions right now as I could not find the call but I may have not found it across the host and webjobs codebase.  
2. By the extension when Dispose() is called.  
3. By the Service bus SDK - I'm not sure under what circumstances.

When we introduced the "drain mode" functionality, the "IListener.StopAsync()" method was used instead of adding a new mechanism.  StopAsync() now comes into play in two circumstances (a) Stopping the functions host while draining all in-flight executions (b) Stopping the functions host quickly where in-flight executions only have as long as asp.net allows services to stop before objects are disposed and the process exits (IIRC its around 15 sec). The customer function will get the cancellation signal when the listener is disposed and will have a bit of time to cleanup before the process exits.  The former happens when scaling in and the latter happens in all other scenarios (e.g. customer stops the app, platform update etc.).  We have spoken about implementing a mechanism in the future for customer code to distinguish between the host draining and the host shutting down quickly but at this time, as there is just one cancellation token, we don't setting as the customer code will have enough time to complete without interruption.

I just checked the [T2 implementation of StopProcessingAsync()](https://github.com/Azure/azure-sdk-for-net/blob/847af3127f7ed0e5c1ff64ce58fe5278eff7f2e9/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs#L734) and unfortunately this behavior was changed from how UnregisterMessageHandlerAsync() works as it cancel's the token passed to the handler :(

To consistent across T1 and T2 in the extension, I feel that [ServiceBusListener.StopAsync()](https://github.com/Azure/azure-sdk-for-net/blob/847af3127f7ed0e5c1ff64ce58fe5278eff7f2e9/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs#L178)  should call ServiceBusProcessor.StopProcessingAsync() instead of ServiceBusProcessor.CloseAsync().  CloseAsync() should be called in the dispose method.  It's not an issue right now though, as we don't restart stopped listeners.

JoshLove-msft: Thank you for that background. It is very helpful.

> I just checked the [T2 implementation of StopProcessingAsync()](https://github.com/Azure/azure-sdk-for-net/blob/847af3127f7ed0e5c1ff64ce58fe5278eff7f2e9/sdk/servicebus/Azure.Messaging.ServiceBus/src/Processor/ServiceBusProcessor.cs#L734) and unfortunately this behavior was changed from how UnregisterMessageHandlerAsync() works as it cancel's the token passed to the handler :(

This is by design - the semantics of the cancellation token are that the token is cancelled when the processor is shutting down. It works this way in both Service Bus and Event Hubs T2 libraries. In any case, I think the change to add this feature was made _after_ the T2 libraries were already GA'ed.

> To consistent across T1 and T2 in the extension, I feel that [ServiceBusListener.StopAsync()](https://github.com/Azure/azure-sdk-for-net/blob/847af3127f7ed0e5c1ff64ce58fe5278eff7f2e9/sdk/servicebus/Microsoft.Azure.WebJobs.Extensions.ServiceBus/src/Listeners/ServiceBusListener.cs#L178) should call ServiceBusProcessor.StopProcessingAsync() instead of ServiceBusProcessor.CloseAsync(). CloseAsync() should be called in the dispose method. It's not an issue right now though, as we don't restart stopped listeners.

I don't think this would have an impact because CloseAsync calls into StopProcessingAsync. In other words, there is no current way to forcefully close the processor other than closing the whole connection. This again was by design as we didn't see a need to support this scenario. Because the cancellation token is signaled, customer code has the option to honor the token and stop whatever they are doing if they need to adhere to some SLA.

alrod:
It looks like we haven't released 4.3.1. Can you please add the fix to list of changes.
https://github.com/Azure/azure-functions-servicebus-extension/blob/dev/README.md

sidkri (approved):

